PR #1, opened 2021-10-27
Update README and add bLIP-0001

ryanthegentry:
This PR initializes the new bLIPs repo with a README modeled after the BIPs [README](https://github.com/bitcoin/bips/blob/master/README.mediawiki), and adds bLIP-0001, which was previously being discussed in PR [#884](https://github.com/lightningnetwork/lightning-rfc/pull/884) in the lightning-rfc repo.

Only a few adjustments have been made to bLIP-0001 while transferring from the previous repo:
1) added a mandated `Universality` section requesting proposals "discuss why the given feature is not intended to be 
universal and why it's still a good idea as a non-universal protocol" per @TheBlueMatt's feedback.
2) added that Feature Bits >100 yet <1000 were reserved for experimentation per @rustyrussell's comments in the [9/27 lightning-dev IRC meeting](https://gnusha.org/lightning-dev/2021-09-27.log), and the same for TLVs > 65536 per [Bolt #1](https://github.com/lightningnetwork/lightning-rfc/blob/master/01-messaging.md#type-length-value-format).
3) fixed a few links/comments that would no longer be correct with a new bLIPs repo (hope I didn't miss any!).

Once these two docs are merged, the next step IMO will be to close the old PR and open a new one for a BOLT #13 (?) that acts as a central registry for bLIP'd Feature Bits, Message Types, and TLVs (and maybe more down the road...) per @t-bast's feedback (replacing PRs [#605](https://github.com/lightningnetwork/lightning-rfc/issues/605) and [#716](https://github.com/lightningnetwork/lightning-rfc/issues/716) from lightning-rfc). Maybe specifying the process of approving a bLIP using one of these resources, and then adding the resource assignment to the new BOLT, could be a good bLIP-0002?


Roasbeef (approved):
Carrying over my approval here from the other repo

LGTM ☄️

TheBlueMatt (approved):

t-bast (approved):

vincenzopalazzo:
Hey Jeff, pushed a fixup commit on `blip-0056-address-review` with everything addressed. The big ones:

- Dropped `order_id` from the protocol entirely (per our Vienna discussion + the Discord thread). Merchant relays the decrypted `payment_token`, PoS matches it against what it sealed. `order_id` is now just PoS-internal data.
- Restructured the whole thing top-down: overview + glossary first, then details.
- Killed all the LDK-specific language.
- Trimmed ~190 lines — it was way too verbose, lots of stuff repeated in 3 places.

Still need to reserve the TLV/feature bit numbers in blip-0002, will do that next. A couple of inline comments below where I would like your input.

PR #4, opened 2021-11-05
Add blip listing reserved values

t-bast:
This is a meta blip that other blips need to update whenever they want to reserve a value in a scarce namespace.
The tables in that blip must be kept up to date to avoid clashes in the network, which would lead to p2p network split.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,5 @@
+# Bitcoin Lightning Improvement Proposal
+
 bLIP stands for Bitcoin Lightning Improvement Proposal. A bLIP is a design document 
 providing information to the Lightning community, or describing a new feature for 
 the Lightning Network. The bLIP should provide a concise technical specification of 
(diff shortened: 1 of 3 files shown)


ryanthegentry:
Two minor nits:
1) Can you please also update the [README](https://github.com/lightning/blips/blob/master/README.md) file to include blip-0002? 
2) Technically blip-0001 says to do numbering by PR number and this is PR #4... should this be blip-0004 instead? I realize that's a bit awkward but don't want to set the wrong precedent...

t-bast:
> Can you please also update the README file to include blip-0002?

Good catch, done and rebased.

> Technically blip-0001 says to do numbering by PR number and this is PR number 4...
> should this be blip-0004 instead? I realize that's a bit awkward but don't want to set the wrong precedent...

As this is a meta blip, I don't think this rule applies and I think it would be confusing.
Ideally it should have been merged at the same time as blip-0001, this is part of the initial setup process imo, we can't accept any "real" blips until we have this table of reserved values somewhere.

ryanthegentry (approved):
LGTM 🥳

PR #8, opened 2022-01-04
Draft of general-purpose Lightning signatures

Kixunil:
This documents LN signatures produced by `signmessage` RPCs of
LND/Eclair/C-Lightning as they are used today.

It was requested by @t-bast at
https://github.com/ACINQ/eclair/issues/2121#issuecomment-1004615247
and documenting seems to be obvious correct thing to do.

Note that I used `##` for headings instead of `#` since [there shouldn't be more than one `#`](https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md025---multiple-top-level-headings-in-the-same-document) and I think other documents should be fixed. However I can change it to `#` if others disagree.

### DIFF
--- a/blip-0008.md
+++ b/blip-0008.md
@@ -54,7 +54,7 @@ Pubkey: `029ef8ee0ba895e2807ac1df1987a7888116c468e70f42e7b089e06811b0e45482` (he
 Message: ``I am the author of `ln-types` Rust crate.`` (ASCII-encoded)
 Signature: `1fdc7b075774fbfbf9b546535f4ff3912f507c134b9e30fa519445a549475228ec71624a6e6671c976837b6f63faf42eed2034c9645cd2b58fd65205794278b01b` (hex-encoded)
 
-Changing any byte in pubkey, message, or signature should produce invalid signature.
+Changing any byte in pubkey, message, or signature should produce invalid signatures.
 
 ## Motivation
 
--- a/blip-0008.md
+++ b/blip-0008.md
@@ -1,5 +1,5 @@
 ```
-bLIP: xxxx
+bLIP: 0008
 Title: General-purpose Lightning signatures
 Status: Draft
 Author: Martin Habovštiak
--- a/blip-0008.md
+++ b/blip-0008.md
@@ -51,7 +51,7 @@ This document does **not** mandate any specific user-facing encoding.
 Valid signature:
 
 Pubkey: `029ef8ee0ba895e2807ac1df1987a7888116c468e70f42e7b089e06811b0e45482` (hex-encoded)
-Message: ``I am the author of \`ln-types\` Rust crate.`` (ASCII-encoded)
+Message: ``I am the author of `ln-types` Rust crate.`` (ASCII-encoded)
 Signature: `1fdc7b075774fbfbf9b546535f4ff3912f507c134b9e30fa519445a549475228ec71624a6e6671c976837b6f63faf42eed2034c9645cd2b58fd65205794278b01b` (hex-encoded)
 
 Changing any byte in pubkey, message, or signature should produce invalid signature.
@@ -81,4 +81,4 @@ Eclair: https://github.com/ACINQ/eclair/pull/1499
 
 C-Lightning: https://github.com/ElementsProject/lightning/pull/3150
 
-`ln-types` Rust library: https://docs.rs/ln-types/0.1.5/ln\_types/node\_pubkey/struct.NodePubkey.html#method.verify\_lightning\_message (verification only)
+`ln-types` Rust library: https://docs.rs/ln-types/0.1.5/ln_types/node_pubkey/struct.NodePubkey.html#method.verify_lightning_message (verification only)
--- a/README.md
+++ b/README.md
@@ -19,6 +19,7 @@ For more detail on the process, please read [bLIP-0001](./blip-0001.md) and
 
 | Number | Title                     | Author                      | Status |
 |--------|---------------------------|-----------------------------|--------|
-| [1](./blip-0001.md)      | bLIP Process              | Ryan Gentry                 | Active |
-| [2](./blip-0002.md)      | reserved values           | Bastien Teinturier          | Active |
-| [3](./blip-0003.md)      | Keysend                   | Valentine Wallace           | Active |
+| [1](./blip-0001.md)      | bLIP Process                         | Ryan Gentry                 | Active |
+| [2](./blip-0002.md)      | reserved values                      | Bastien Teinturier          | Active |
+| [3](./blip-0003.md)      | Keysend                              | Valentine Wallace           | Active |
+| [8](./blip-0008.md)      | General-purpose Lightning signatures | Martin Habovštiak           | Draft  |
--- a/blip-xxx.md
+++ b/blip-xxx.md
@@ -1,0 +1,84 @@
+```
+bLIP: xxxx
+Title: General-purpose Lightning signatures
+Status: Draft
+Author: Martin Habovštiak
+Created: 2022-01-04
+License: CC0
+```
+
+## Abstract
+
+This document describes general-purpose signatures tied to Lightning Network
+node identifiers (pubkeys) that can be used in user-facing applications or
+custom extension protocols ("3rd layer") e.g. to implement reputation or
+advertising of LN nodes. They are **not** intended to be used
+in the Lightning Network Protocol itself (not BOLT).
+
+This bLIP serves to document what is already well-supported in the wild (see
+[Reference Implementations](#reference-implementations)), for posterity and so
+that new implementations don't have to reverse-engineer general-purpose
+Lightning signatures from existing implementations.
+
+## Copyright
+
+This bLIP is licensed under the CC0 license.
+
+## Specification
+
+The signature is ECDSA secp256k1 recoverable signature using *double* sha256 as
+the hasing algorithm. The message MUST be prefixed with ASCII-encoded text
+`Lightning Signed Message:` before signing/verification.
+The content or encoding of the message is unrestricted.
+
+The signature is encoded as 65 bytes: one-byte recovery ID and 64 bytes of
+compact signature as defined by `libsecp256k1`:
+
+> The signature must consist of a 32-byte big endian R value, followed by a
+> 32-byte big endian S value. If R or S fall outside of [0..order-1], the
+> encoding is invalid. R and S with value 0 are allowed in the encoding.
+
+The constant `0x1f` has to be added to the recovery ID before encoding.
+(e.g. recovery ID 0x00 will cause the final signature begin with 0x1f)
+
+Multiple implementations encode the signature in zbase32 when presenting it to
+the user. At the time of writing Eclair uses hex instead because zbase32 is
+uncommon.
+This document does **not** mandate any specific user-facing encoding.
+
+### Test Vectors
+
+Valid signature:
+
+Pubkey: `029ef8ee0ba895e2807ac1df1987a7888116c468e70f42e7b089e06811b0e45482` (hex-encoded)
+Message: ``I am the author of \`ln-types\` Rust crate.`` (ASCII-encoded)
+Signature: `1fdc7b075774fbfbf9b546535f4ff3912f507c134b9e30fa519445a549475228ec71624a6e6671c976837b6f63faf42eed2034c9645cd2b58fd65205794278b01b` (hex-encoded)
+
+Changing any byte in pubkey, message, or signature should produce invalid signature.
+
+## Motivation
+
+These signatures can be useful for proving ownership of an LN node, initiating
+secure communication between node operators (e.g. by signing GPG keys with node
+pubkeys) or other purposes.
+
+As mentioned earlier this bLIP describes what already exists and is being used
+and supported by multiple popular LN implementations, so it's useful to document it.
+
+## Rationale
+
+Double hashing and prefixing the message with a constant text should minimize
+the risk of these signatures being confused with invoices or other LN-related
+signatures. The implementors of higher-level protocols may want to implement
+additional measures to prevent confusion or replay attacks among multiple
+higher-level protocols.
+
+## Reference Implementations
+
+LND: https://github.com/lightningnetwork/lnd/pull/192
+
+Eclair: https://github.com/ACINQ/eclair/pull/1499
+
+C-Lightning: https://github.com/ElementsProject/lightning/pull/3150
+
+`ln-types` Rust library: https://docs.rs/ln-types/0.1.5/ln\_types/node\_pubkey/struct.NodePubkey.html#method.verify\_lightning\_message (verification only)


blip-0008.md
```diff
@@ -0,0 +1,84 @@
...
+## Copyright
+
+This bLIP is licensed under the CC0 license.
+
+## Specification
+
+The signature is ECDSA secp256k1 recoverable signature using *double* sha256 as
+the hasing algorithm. The message MUST be prefixed with ASCII-encoded text
+`Lightning Signed Message:` before signing/verification.
+The content or encoding of the message is unrestricted.
```
t-bast, line 32: Can you replace that by pseudo-code (with a small introduction saying that it's basically an ECDSA signature of a tagged hash of the message)?
That would be much easier to read and implement.

blip-0008.md
```diff
@@ -0,0 +1,84 @@
...
+
+The signature is encoded as 65 bytes: one-byte recovery ID and 64 bytes of
+compact signature as defined by `libsecp256k1`:
+
+> The signature must consist of a 32-byte big endian R value, followed by a
+> 32-byte big endian S value. If R or S fall outside of [0..order-1], the
+> encoding is invalid. R and S with value 0 are allowed in the encoding.
```
t-bast, line 39: I would link to some kind of specification for ECDSA compact sigs instead of copy-pasting this, it feels out of place as it references things that aren't defined here (e.g. the order of the curve).

blip-0008.md
```diff
@@ -0,0 +1,84 @@
...
+
+The constant `0x1f` has to be added to the recovery ID before encoding.
+(e.g. recovery ID 0x00 will cause the final signature begin with 0x1f)
```
t-bast, line 42: I don't remember the reason we do this. @Roasbeef do you have a clue?
It would be useful to explain it here.

blip-0008.md
```diff
@@ -0,0 +1,84 @@
...
+
+### Test Vectors
+
+Valid signature:
+
+Pubkey: `029ef8ee0ba895e2807ac1df1987a7888116c468e70f42e7b089e06811b0e45482` (hex-encoded)
+Message: ``I am the author of `ln-types` Rust crate.`` (ASCII-encoded)
+Signature: `1fdc7b075774fbfbf9b546535f4ff3912f507c134b9e30fa519445a549475228ec71624a6e6671c976837b6f63faf42eed2034c9645cd2b58fd65205794278b01b` (hex-encoded)
+
+Changing any byte in pubkey, message, or signature should produce invalid signature.
```
t-bast, line 57: nit:
```suggestion
Changing any byte in pubkey, message, or signature should produce invalid signatures.
```

t-bast (commented):
Thanks! You should also add a link to the main `README.md`.

> Note that I used ## for headings instead of # since there shouldn't be more than one # and I think other documents should be fixed.

I agree with you. Will fix the other bLIPs in a separate PR.

It would be useful to have a review from one of the lnd devs (@Roasbeef can you review this or assign someone?), since they implemented this first and made the design choices.

blip-0008.md
Kixunil: Is this notation OK or you have a better idea?

`secp256k1_ecdsa_sign(key, sha256(sha256("Lightning Signed Message:" + message)))`

blip-0008.md
Kixunil: AFAIU ECDSA compact sigs are only "specified" by `libsecp256k1` code. `R`, `S` and `order` seemed self-evident but I may be wrong.

TheBlueMatt (commented):
This is missing a section on why this isn't targeted for general applicability (ie why a bLIP and not a BOLT), and I actually think we should Just Land this as a new BOLT, its not like we have to do cross-implementation compat testing for it or anything, just two ACKs and we can merge.

Kixunil:
I think @t-bast knows better than me since he suggested this.

blip-0008.md
t-bast: Yes, I think this is detailed enough and understandable by anybody working on a lightning implementation.
You could expand it slightly:

```
* signedBytes := sha256(sha256("Lightning Signed Message:" + message))
* signature := secp256k1_ecdsa_sign(node_private_key, signedBytes)
```

blip-0008.md
t-bast: What about just linking to that page: https://en.bitcoin.it/wiki/Elliptic_Curve_Digital_Signature_Algorithm ?

t-bast:
I think it shouldn't be a BOLT because it's not necessary for the lightning protocol at all.
It's only used for applications outside of the core of lightning AFAIK.

Roasbeef:
> I think it shouldn't be a BOLT because it's not necessary for the lightning protocol at all.

Yeah agreed, the base protoco never sees this, given it only exists on the API/RPC layer for a given node implementation.

blip-0008.md
Roasbeef: This is part of the compact signature encoding: a value is added to signify the iteration count/index that should be used to recover the signature (as there's 4 diff combos).

blip-0008.md
Roasbeef: The optimization that's somewhat Bitcoin-specific here is adding the recovery byte, so the public key can be recovered in a single iteration. See the comment below re the recovery byte.

blip-0008.md
Roasbeef: So instead, this should specify that this constant is the recovery byte, and will depend on which of the 4 iteration result in proper public key recovery. Also fwiw, this is how "bitcoin signed message"s work as well.

Roasbeef (changes requested):

TheBlueMatt:
> I think it shouldn't be a BOLT because it's not necessary for the lightning protocol at all.

Is that really the criteria for bLIPs? I thought we had a rough handle on it with the "not intended to be cross-compatible" definition. If we move to a "is it in the core protocol" definition we might as well move all invoice definitions to bLIPs, but I don't think that makes much sense.

Roasbeef:
> If we move to a "is it in the core protocol" definition we might as well
move all invoice definitions to bLIPs, but I don't think that makes much
sense.

Put another way: a shared invoice format is a requirement to be able to
properly send+receive payments across implementations. Hard to test interop
w/o a shared invoice format: it's a critical part of a working network and
_everyone_ needs to agree on it.

This document specifies a way to validate signatures (bound to the LN
context w/ that message prefix) generated by nodes that's _already_ being
used by some implementations. As this is purely client/wallet-side, this
doesn't require broad consensus, and imo it's preferable not to add that
requirement as that slows down the nimbleness of the scheme when people
want to change it in the future.

IMO, we should just let the authors pick which type of document they want
to write. Any sort of justification they give is enough, as at least they
have a home to specify things they've already done that don't require broad
consensus/adoption.

t-bast:
> I thought we had a rough handle on it with the "not intended to be cross-compatible" definition.

I understood a slightly different version, something more along the lines of "not intended to be implemented by everyone". I believe this PR does fit that criteria: it's not necessary at all for a lightning implementation to support signing/verifying arbitrary messages, is it?

Also, if we were adding signing/verifying messages to the BOLTs, I believe we'd like to discuss the protocol and could end up with something slightly different from this one. Since this blip only specifies what already exists, I think it brings some value.

TheBlueMatt:
bLIP 1 currently states that each bLIP must have a section on

> Universality -- This section should discuss why the given feature is not intended to be universal and why it's still a good idea as a non-universal protocol. New features intended to be universally deployed should go through the BOLTs process instead.

that's pretty clear IMO, and this doesn't fall into that category (and, really, isn't worth debating too much, because there shouldn't be any issue with getting it merged as a BOLT within one meeting - there's already lots of implementations :) ). Indeed, if the author wishes to write a section on why this spec isn't intended to be universally deployed, it could be merged as a bLIP, but we had this debate when bLIP 1 was written, we can reopen it as well, but that would likely take longer than just merging this as a BOLT.

ryanthegentry:
Curious, do any implementations use this RPC for in-protocol things, or is it just available for applications to use? @TheBlueMatt you're suggesting this should be a new standalone BOLT doc, not folded into an existing one, right?

TheBlueMatt:
> Curious, do any implementations use this RPC for in-protocol things, or is it just available for applications to use?

I'm not aware of any such use, I do believe its an exposed-to-application/users thing, hence, IMO, why there should be compatibility around it :).

> @TheBlueMatt you're suggesting this should be a new standalone BOLT doc, not folded into an existing one, right?

I assume that would be the simplest approach, yes, I don't see why it would fit into any other BOLTs and it seems like a good candidate for its own one.

ryanthegentry:
Hmm... given that this signmessage already exists in the wild across all implementations, I don't think it qualifies as a "New feature intended to be universally deployed". Seems like a perfect example instead of an "app layer best practice to be succinctly described and organized, especially [one] that require[s] coordination" as described in [the original bLIP email](https://lists.linuxfoundation.org/pipermail/lightning-dev/2021-June/003086.html). Very possible bLIP-0001 is insufficiently capturing that nuance, mea culpa if so and happy to address. 

Personally I think we should applaud @Kixunil for documenting this, maybe add a "Universality" section talking about how this is documenting an existing feature instead of proposing a new one, merge the doc, and then have a separate discussion about whether this signmessage RPC is really a Basic of Lightning Technology given it isn't used for any in-protocol Lightning things.

Kixunil:
Personally I don't care much. It seems that neither is good but I feel bLIP is less bad.

TheBlueMatt:
There was a discussion on this in the meeting today, after quite some back-and-forth, a suggested path forwards emerged that had some reasonable level agreement. (a) define a BOLT on the sign and verify message format itself, but not the human-readable parts. The BOLT should suggest hex-encoded for human readable format signatures to avoid pulling in a new zbase32 dependency into BOLTs. (b) document the existing deployment of zbase32 in a bLIP, noting that implementations MUST accept hex-encoded signatures and MAY accept zbase32 for backwards compatibility. Some discussion of migrating to hex-encoded signatures on the creation end could also be included.

If you're willing to write both, @Kixunil, that would be much appreciated, but otherwise I or someone else may be willing to do the other parts of it.

Kixunil:
Ah, OK, I guess I will start with BOLT, pretty much copying what was done here with the requested changes and then perhaps open a new PR here. Although TBH just saying pretty much "some implementations use zbase32 but ideally all should use hex" seems quite short for any such document. Maybe a reason to not care much and just do it though. :D

TheBlueMatt:
Short documents are best documents IMO :). No reason we aren't allowed to have short documents if they're describing something important about how lightning operates in the field today :)

Roasbeef:
> Personally I don't care much. It seems that neither is good but I feel bLIP is less bad.

As it's already something that's in the wild and "emerged" over time, a bLIP is the correct place. 

Making something like this a BOLT is just attempting to shove even more things into what's already a pretty review constrained process. BOLTs are good for things that require deep require and wide spread consensus like integrating PTLCs into the routing network. Defining an _application level_ signature system, that isn't in use in the **anywhere** _base_ protocol is outside of the purview of BOLTs. 

I think @TheBlueMatt's stance here represents a backwards attitude that pushes would be contributors away from participating in either the bLIP or BOLT process. People just want to make cool things, and occasionally write them up for others to check out. We should be more inviting for people to participate at w/e level they're inclined to do so. By the end of the year we'll likely have many more individual bLIP documents than BOLT documents (given the lower barrier to entry, and expanded scope). Most of them will likely have a short PR lifetime compared to BOLT proposals (due to the differences in the processes), which is great as it allows the maintainers of the existing implementations to focus their attention on larger scope/impact proposals in the BOLT arena. I know it's hard, but sometimes it's OK to just not have a strong opinion on something.

TheBlueMatt:
> Making something like this a BOLT is just attempting to shove even more things into what's already a pretty review constrained process.

I think you've hit the nail on the head here, but with the wrong solution - the BOLT process' review speed has improved greatly over the past months, and I'm looking forward to it continuing to improve as ACINQ, Blockstream, Lightning Labs, and Spiral are able to hire going into this new year. In this specific discussion, I took pains to point out that, given its already broadly deployed and compatibility-tested, I see absolutely no reason why merging a BOLT which describes this protocol cannot happen within a few weeks, no more or less fast than merging a bLIP. Given there's already rough agreement of the form of it from the meeting on monday, that seems to be likely to be true.

All that said, I've been very supportive of the bLIP process since the beginning, please don't take any comments about a specific PR as somehow a blocker to bLIPs existing, that just feels like an argument ad absurdum. When the bLIP process was proposed there were many questions about how lightning UX will evolve as more protocols and optional features exist, and we agreed to strike a balance that I echoed here. Just because something arguably does or does not fit into that balance doesn't mean I'm somehow against bLIPs.

More importantly, please of course note that the bLIP 1 doc suggests that a bLIP should have a discussion section about the very balance that we struck, which was missing in this document. I assume if it were to have been written even the author would have found it lacking, hence why I suggested that it may be *faster* to simply write a BOLT which could be merged within a week or two, rather than trying to twist words to make it so that this can be taken as a bLIP.

Kixunil:
Could we perhaps create a third repository called "Lightning Documentation Dumping Area" (LDDA)? Basically bLIPs with "not intended to be universal" requirement replaced by "not protocol directly communicated by LN nodes". That way Lightning Developers Don't need to Argue about these things and we can move forward... :)

TheBlueMatt:
Sadly, that was precisely part of the argument about whether to have bLIPs or not - lightning's UX has begin to suffer quite substantially from lack of compatibility across nodes (see, for example, the horrendous UX of 0-amount invoices depending on the client). There is of course a line to walk - experimentation is critical to lightning continuing to grow into new use-cases and domains, but so is UX, a part of which is compatibility across lightning clients and wallets.

I don't think we're going to settle this issue here, and part of the discussion on bLIPs was that ultimately this is the kind of thing that is going to have to be done on a case-by-case basis. In this case we arguably have a pretty easy way forward that everyone (but @Roasbeef, apparently) seemed happy with in the meeting on Monday (it was a holiday in the US so @Roasbeef could not attend), so I don't think its necessarily worth reopening that debate here, in any case.

ryanthegentry:
>Could we perhaps create a third repository called "Lightning Documentation Dumping Area" (LDDA)? Basically bLIPs with "not intended to be universal" requirement replaced by "not protocol directly communicated by LN nodes"

This is basically what the bLIPs are intended to be, and why we agreed on Monday that this should proceed as a bLIP if @Kixunil wouldn't mind adding a brief "Universality" section that states "This isn't a candidate for universality because zbase32 is [used by some implementations and not others, and not a dependency we want to build into the spec]" (paraphrasing from [the notes](https://gnusha.org/lightning-dev/2022-01-17.log) :)).

The BOLT step specifying an ideal way to sign messages extra-protocol is separate, if protocol devs want to do devote resources to writing it and cross implementation testing and all that jazz. I don't think this PR should've been closed, I think it needs the aforementioned sentence included and should be merged.

TheBlueMatt:
> and why we agreed on Monday that this should proceed as a bLIP if @Kixunil wouldn't mind adding a brief "Universality" section

To be fair, there's also nothing wrong with the OP here proposing it as a BOLT as discussed as well, which is presumably why this was closed.

I'm still somewhat confused why we all seem to be treating "proposing a BOLT" as some huge thing that has insurmountable cost, especially given we've substantially improved BOLT review and merge time over the past few months, largely merging things within a few weeks, after cross-implementation testing completes (something already done here).

t-bast:
FWIW, I would spend the time to review a BOLT for this feature and help it get integrated quickly.
However, if there is strong push-back from other maintainers on having this as a BOLT, I'd also spend the time to review it as a bLIP if necessary.

So my simple proposal is to simply open a PR on the BOLTs repo, try to get it integrated quickly, and if that fails we'll fall back to making it a bLIP.

Kixunil:
> I don't think this PR should've been closed, I think it needs the aforementioned sentence included and should be merged.

Hmm, @TheBlueMatt says it should be BOLT and this one should only describe encoding which I felt is so different from the current text it'll be cleaner to open a new one (after merging one against BOLT). So I went back to see the notes and indeed people agreed with @TheBlueMatt so I'll do it this way.

PR #10, opened 2022-01-25
Podcasting 2.0 blip

satoshisstream:
_Thanks Lucas from LL helping drafting the blip_

Based on the TLV registry: https://github.com/satoshisstream/satoshis.stream/blob/main/TLV_registry.md please find a formalized blip in this PR. 

Let me know if I need to do something!

### DIFF
--- a/blip-0002.md
+++ b/blip-0002.md
@@ -93,15 +93,8 @@ The following table contains extension tlv fields for the `update_add_htlc` mess
 
 | Type       | Name                        | Link                           |
 |------------|-----------------------------|--------------------------------|
-| 5482373484 | `keysend_preimage`          | [bLIP 3](./blip-0003.md)       |
-
-#### `podcasting_2_0`
-
-The following table contains extension tlv fields for the `update_add_htlc` message's onion payload:
-
-| Type       | Name                        | Link                           |
-|------------|-----------------------------|--------------------------------|
 | 7629169    | `podcasting_2_0`            | [bLIP 10](./blip-0010.md)       |
+| 5482373484 | `keysend_preimage`          | [bLIP 3](./blip-0003.md)       |
 
 ## Copyright
 
--- a/blip-0002.md
+++ b/blip-0002.md
@@ -79,6 +79,15 @@ The following table contains extension tlv fields for the `init` message:
 |-------|-----------------------------|--------------------------------|
 | 65536 | `tlv_field_name`            | Link to the corresponding bLIP |
 
+#### `payment_onion_payload`
+
+The following table contains extension tlv fields for the `payment_onion_payload` message:
+
+| Type        | Name                        | Link                           |
+|-------------|-----------------------------|--------------------------------|
+| 7629169     | `podcasting_2_0`            | [bLIP 10](./blip-0010.md)      |
+
+
 #### `ping`
 
 The following table contains extension tlv fields for the `ping` message:
@@ -93,7 +102,6 @@ The following table contains extension tlv fields for the `update_add_htlc` mess
 
 | Type       | Name                        | Link                           |
 |------------|-----------------------------|--------------------------------|
-| 7629169    | `podcasting_2_0`            | [bLIP 10](./blip-0010.md)       |
 | 5482373484 | `keysend_preimage`          | [bLIP 3](./blip-0003.md)       |
 
 ## Copyright
--- a/blip-0002.md
+++ b/blip-0002.md
@@ -95,6 +95,14 @@ The following table contains extension tlv fields for the `update_add_htlc` mess
 |------------|-----------------------------|--------------------------------|
 | 5482373484 | `keysend_preimage`          | [bLIP 3](./blip-0003.md)       |
 
+#### `podcasting_2_0`
+
+The following table contains extension tlv fields for the `update_add_htlc` message's onion payload:
+
+| Type       | Name                        | Link                           |
+|------------|-----------------------------|--------------------------------|
+| 7629169    | `podcasting_2_0`            | [bLIP 10](./blip-0010.md)       |
+
 ## Copyright
 
 This bLIP is licensed under the CC0 license.
--- a/blip-0010.md
+++ b/blip-0010.md
@@ -22,6 +22,19 @@ This bLIP is licensed under the CC0 license.
 
 The sender of the payments (the Podcast app) needs to input some metadata so the podcast host can identify whom the payment is for. Most fields are optional. 
 
+### Formatting and encoding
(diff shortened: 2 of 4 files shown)


lucasdcf:
🚀🚀🚀

ryanthegentry:
Hi @satoshisstream, thanks so much for documenting! Some initial housekeeping, could you please:

1) rename the doc to blip-0010, given Github's numbering convention
2) include in this PR an update to [blip-0002](https://github.com/lightning/blips/blob/master/blip-0002.md#tlv-fields-in-bolt-messages) "registering" TLV type `7629169` with name `Boostagram` (?) and a link to the future blip-0010? I assume this is an `update_add_htlc` message like keysend, so [can follow that PR](https://github.com/lightning/blips/pull/5) as an example if you'd like.
3) include in this PR an update to the [README](https://github.com/lightning/blips/blob/master/README.md) adding blip-010 (again following the keysend bLIP as an example).

satoshisstream:
Thanks!

How about that? :)

blip-0002.md
```diff
@@ -95,6 +95,14 @@ The following table contains extension tlv fields for the `update_add_htlc` mess
 |------------|-----------------------------|--------------------------------|
 | 5482373484 | `keysend_preimage`          | [bLIP 3](./blip-0003.md)       |
 
+#### `podcasting_2_0`
+
+The following table contains extension tlv fields for the `update_add_htlc` message's onion payload:
```
t-bast, line 100: That means you can only pay someone you're directly connected to, is that right?
You cannot use this with multi-hop payments?

blip-0002.md
```diff
@@ -95,6 +95,14 @@ The following table contains extension tlv fields for the `update_add_htlc` mess
 |------------|-----------------------------|--------------------------------|
 | 5482373484 | `keysend_preimage`          | [bLIP 3](./blip-0003.md)       |
 
+#### `podcasting_2_0`
```
t-bast, line 98: Don't create a new section, there is one section per lightning message.
You should add your TLV to the table in the `update_add_htlc` section.

blip-0010.md
```diff
@@ -0,0 +1,86 @@
...
+Status: Active
+Author:  Satoshis Stream <satoshisstream@pm.me>
+Created: 2022-01-24
+License: CC0
+```
+
+## Abstract
+
+Using Podcasting 2.0 apps, listeners can send sats to the hosts of the podcasts they listen. Per minute sending is called `stream` and manual payments are called `boost`s. If there is a message to the podcaster, with a boost, it is called a `boostagram`.
+This is part of what is known as "the Value for Value system". These apps (see [Reference Implementations](#reference-implementations)) adopted the 7629169 TLV type for inputting key-value JSON metadata about the sent payment. The TLV holds data about the timestamp when the payment was sent within the episode, the name of the podcast and its [Guid](https://github.com/Podcastindex-org/podcast-namespace/blob/main/docs/1.0.md#guid), among other fields described in the [Specification](#Specification) section.
```
t-bast, line 13: This is under-specified: TLVs use a binary encoding.
You must clarify how you encode your JSON object into binary.
Do you just treat it as a utf8 string? ASCII?
JSON ordering isn't deterministic, so you should be aware that it may create issues.

blip-0010.md
```diff
@@ -0,0 +1,86 @@
...
+## Copyright
+
+This bLIP is licensed under the CC0 license.
+
+## Specification
+
+The sender of the payments (the Podcast app) needs to input some metadata so the podcast host can identify whom the payment is for. Most fields are optional. 
+
+Identifying the podcast **required**: use any of `podcast`, `feedID` or `url`. **guid preferred**
+* `podcast` (str) Title of the podcast
```
t-bast, line 26: Wherever you say `str` here you need to clarify its binary encoding.
It would be very useful to have test vector where you would show the JSON and the corresponding encoded tlv.

blip-0002.md
satoshisstream: Done

blip-0002.md
satoshisstream: It's regular keysend so the sender (podcast app) would just find a route and pay :). Works well!

t-bast: Then this is incorrect, this isn't a TLV on the `update_add_htlc` message, this is a TLV in the onion?

blip-0010.md
satoshisstream: UTF8 indeed. Added info.

satoshisstream:
Thank you @t-bast . I believe I have addressed all your good comments.

TheBlueMatt (approved):
Thanks for taking the time to write this up!

blip-0010.md
```diff
@@ -0,0 +1,99 @@
...
+## Copyright
+
+This bLIP is licensed under the CC0 license.
+
+## Specification
+
+The sender of the payments (the Podcast app) needs to input some metadata so the podcast host can identify whom the payment is for. Most fields are optional. 
+
+### Formatting and encoding
+A flat key-value json structure is used where below keys can be set. The string is then encided into `utf8` and attached to the keysend payment. Receivers of messages must be aware there is no guarantee for the order of the keys.
```
t-bast, line 26: nit: typo:
```suggestion
A flat key-value json structure is used where below keys can be set. The string is then encoded into `utf8` and attached to the keysend payment. Receivers of messages must be aware there is no guarantee for the order of the keys.
```

t-bast (commented):
Thanks! I haven't verified the test vector myself though.

There is something that is still unclear to me, related to this comment: https://github.com/lightning/blips/pull/10#discussion_r791966660

If you want this to work without being directly connected to the person you pay, your TLV shouldn't be in the `update_add_htlc` namespace, it should be in the `payment_onion_payload` namespace (for which you should create a new section in blip 2)?

Do note that this heavily restricts the amount of data available, as onions have a small fixed size.

satoshisstream:
Added the `payment_onion_payload` part. For blip2 I just follow what you say, the chapters are a bit unclear to me. Hope I did it correctly ;)

t-bast:
> Added the payment_onion_payload part. For blip2 I just follow what you say, the chapters are a bit unclear to me. Hope I did it correctly ;)

Thanks, that looks good enough.

Unfortunately we don't have a great way to reference the onion used inside `update_add_htlc` (the bolts don't explicitly name it) but I think it should be easy enough to understand for readers (we do want to distinguish onions used in `update_add_htlc` from onions used in onion messages as they don't use the same tlvs).

Looks good to me now :+1:

t-bast (approved):

ryanthegentry (approved):
LGTM 🎉

Issue #12, opened 2022-01-29
bLIP-10: ts property needs to specify time units


coldacid:
While we may all know right now that `ts` is supposed to be in seconds, there's nothing actually specifying that in the document. Someone could look at it and think that perhaps it's milliseconds. Or perhaps something odd, like the number of frame headers passed in the mp3 file. Long story short, the time unit should be specified explicitly in the doc.

JssDWt:
This is fixed in https://github.com/lightning/blips/pull/13, so can be closed?

satoshisstream:
yes

SiddheshKukade:
Hello @satoshisstream , what does the `ts` field stands for here 
does it mean time in seconds? Please explain.
![image](https://user-images.githubusercontent.com/65951872/231345635-5f0ffa13-68f0-4708-9fe7-c619be02fae6.png)
Source: https://github.com/lightning/blips/blob/master/blip-0010.md

PR #13, opened 2022-01-31
Timestamp is in seconds

satoshisstream:
Define timestamp is in seconds.

https://github.com/lightning/blips/issues/12


coldacid (approved):

TheBlueMatt (approved):
Just a clarification, and authors can update their own docs.

PR #15, opened 2022-02-04
keysend: correct TLV extension field table

valentinewallace:
Keysend's TLV extension type was documented the wrong (and indeed bogus) table, as pointed out by @joostjager [here](https://github.com/lightning/blips/pull/5#discussion_r799291268)


t-bast (approved):
LGTM, thanks!

PR #16, opened 2022-05-28
LNURL-withdrawPOS flow 

theDavidCoen:
Pull request for blip regarding LNURL-withdrawPOS flow standard.


blip-0015.md
```diff
@@ -0,0 +1,96 @@
...
+```
+
+## Abstract
+
+LNURL-withdraw is a bech32-encoded HTTPS query string or sub-protocol of LNURL 
+which gives the users the ability to receive funds via Lightning Network without the need of manually create an invoice.
+Users scan a QRcode or paste the LNURL-withdraw link into their wallet, 
+this queries a server and gets a JSON with some info, 
+such as the max amount the user can receive, the min amount the user can request, etc. 
+Then the wallet tipically asks the user to enter an amount and under the hood it creates a Lightning invoice and sends it to the service provider,
```
cdecker, line 17:
```suggestion
Then the wallet typically asks the user to enter an amount and under the hood it creates a Lightning invoice and sends it to the service provider, 
```

cdecker: Maybe mention that the invoice creation MUST respect the parameters returned by the server.

blip-0015.md
```diff
@@ -0,0 +1,96 @@
...
+## Abstract
+
+LNURL-withdraw is a bech32-encoded HTTPS query string or sub-protocol of LNURL 
+which gives the users the ability to receive funds via Lightning Network without the need of manually create an invoice.
+Users scan a QRcode or paste the LNURL-withdraw link into their wallet, 
+this queries a server and gets a JSON with some info, 
+such as the max amount the user can receive, the min amount the user can request, etc. 
+Then the wallet tipically asks the user to enter an amount and under the hood it creates a Lightning invoice and sends it to the service provider, 
+which eventually pays that.
+So the only things the users have to do are:
```
cdecker, line 19:
```suggestion
From a high level the user-visible flow consists of:
```

blip-0015.md
```diff
@@ -0,0 +1,96 @@
...
+
+1. Scan the QR code / paste the LNURL-withdraw link into the wallet
+2. Enter the amount to receive
+3. Confirm
+
+In this repo I schematize a LNURL-withdraw flow for POS,
```
cdecker, line 25: Don't use the first person to write a specification, use neutral language.

blip-0015.md
```diff
@@ -0,0 +1,96 @@
...
+Then the wallet tipically asks the user to enter an amount and under the hood it creates a Lightning invoice and sends it to the service provider, 
+which eventually pays that.
+So the only things the users have to do are:
+
+1. Scan the QR code / paste the LNURL-withdraw link into the wallet
+2. Enter the amount to receive
+3. Confirm
+
+In this repo I schematize a LNURL-withdraw flow for POS, 
+which sees the interaction between the user and a POS equipped with a NFC sensor.
```
cdecker, line 26: NFC in this case seems more like an example transport rather than an explicit limitation.

blip-0015.md
```diff
@@ -0,0 +1,96 @@
...
+
+This bLIP serves to document what could be a reference flow for open source implementations of POS devices able to read LNURL-withdraw links via NFC,
```
cdecker, line 28:
```suggestion
This bLIP serves to documents the flow for open source implementations of POS devices able to read LNURL-withdraw links via NFC,
```

Again, NFC seems more like a detail here, artificially limiting the applicability (links, QR codes, etc are just as valid as transports).

blip-0015.md
```diff
@@ -0,0 +1,96 @@
...
+for a better UX in Lightning payments powered by LNURL protocol.
```
cdecker, line 29: `better` as compared to what?

blip-0015.md
```diff
@@ -0,0 +1,96 @@
...
+With this flow the users can also make "offline" payments and this completely change the UX of a Lightning payment: 
+they could reserve a part of their funds on a LN wallet for "offline" payments in a LNURL-withdraw string, 
+so they can use it offline, for example when they are abroad and don't have a stable connection or have roaming issues. 
+They could also define sub-balances in their wallet, so once the reserved funds are ended, no other funds in their wallet is touched.
+
+![LNURL-withdraw POS offline payments](https://raw.githubusercontent.com/theDavidCoen/blips/master/blip-0015/offline%20payments%20LNURL-withdrawPOS.png)
+
+## Rationale
+
+There are security concerns in regards to LNURL-withdraw links used as prepaid Lightning cards, which are expressed [here](https://github.com/theDavidCoen/LNURL-withdrawPOS#security-concerns-in-regards-to-lnurl-withdraw-links), but this bLIP aims just to clarify how the receiving part of the flow works and to represent a guide for an open standard, so it's not its goal to highlight them.
```
cdecker, line 60: It is better to document limitations such as these security concerns in the proposal itself, rather than linking to external resource which may change over time, and links may break. In particular security concerns and caveats should be well-documented here, as implementors are unlikely to ever read them if not readily accessible.

cdecker (changes requested):
Very nice @theDavidCoen, documenting this protocol is definitely overdue, so thanks for picking up the slack.

I went through and commented whenever something jumped at me, some of the comments are more pedantic than necesary so feel free to ignore, but I wanted to give concrete examples of how I'd written it (doesn't mean that mine would have been any more correct) :-)

theDavidCoen:
Ehy @cdecker thanks a lot for your comments and suggestions!
It's my first PR here so I really appreciate your help.
I merged some of them and edited the text as well, however I have a couple of questions:
1. Lines 23-26, can you clarify what you mean with "NFC in this case seems more like an example transport rather than an explicit limitation." ?
2. Rationale - I documented the security concerns, can you take a look?
3. A Github related question since I'm a newbie: can you help me changing the number of the blip? I can change the number inside the text and of the .md but I'm not able to change the number of the folder.
I believe the correct number should be blip16 since the PR is #16 .

cdecker:
Ciao @theDavidCoen.

1. It seems to me like there is no requirement to limit ourselves to MFC or qr codes, and that allowing other methods of transferring the LNurl details would be fine, so why explicitly limit yourself to NFC?

2. Will take a look 👍

3. That's more of a git question than GitHub specific, but the way you'd do it is using `git mv oldname newname` and then commit the changes

theDavidCoen:
@cdecker sorry for the late reply.
Please note that this blip doesn't want to document LNURL protocol itself, but a specific flow for NFC payments using LNURL-withdraw.
As mentioned in the description, there are already implementations using a similar (or even identical) flow, and the goal is to have them and future solutions using a standard flow such as this one.

GeorgeTsagk:
Isn't this spec heavily dependent on the LNURL specifications?

- Why is this a PR to the **BLIPs** repository? (and not a PR to `fiatjaf/lnurl-rfc` ?)
- If LNURL related specs were to be introduced within the namespace of BLIPs, shouldn't the whole set of LNURL specifications be migrated here? There is no exact definition of what the responsibilities of BLIPs are, but if we want to include anything LNURL related I believe the whole LNURL umbrella should be introduced for consistency.

theDavidCoen:
@GeorgeTsagk thanks for your comment.
In regards to your questions:
1. This is a UX flow, not a new LNURL sub-protocol.
In fact, it uses LUD 3 (but can use other LNURL-w sub-protocols).
2. I agree that LNURL specs should be also added here as a blip.

lucasdcf:
Chiming in to agree with David that the LNURL specs could totally be a set of blips (one for each LUD)

cdecker:
I'd like this, and the other LNURL specs, to be mirrored here, since they are being adopted by a large number of implementations, wallets and services, and are clearly part of the LN ecosystem. It'd serve as an acknowledgement and would facilitate finding them by developers looking for solutions to their problems. 

Notice that in this repository we only perform stylistic and understandability checks, not discussions on the merits or design choices in order not to dissuade users from pursuing various solutions, as such there would be no barrier to mirroring LNURL, and you might get some feedback on whether the documentation is complete and understandable (I too sometimes take things as given, but they are only because I was present while they were discussed, and someone joining later might not have that context).

This is however completely up to the LNURL team to decide.

PR #18, opened 2022-08-29
Inbound routing fee discount

joostjager:
The Lightning BOLTs define a routing fee that is based on the policy of the _outgoing_ channel. This makes it impossible for routing node operators to differentiate in fees between incoming channels.

Incoming traffic from some peers may be more desired than from others, for example because it balances out traffic in the opposite direction.

This bLIP defines an optional `channel_update` message field to express an inbound fee discount. Senders who are able to read this field can benefit from lower routing fees using those channels.

#### Alternative: pairwise fees

An alternative scheme for inbound fees are the so-called pairwise fees, where a distinct fee can be set for each combination of incoming and outgoing channel. This can also be implemented as a discount to remain backwards compatible.

For this BLIP, I chose to start simple with a plain single inbound fee. A single fee only requires trivial changes to pathfinding algorithms and there is no potential for quadratic growth of gossip data.

If pairwise fees prove to be required, an additional BLIP can be defined. It can again be formulated as another discount on the general inbound discount, to keep all options open for implementers.

### Additional information
* https://github.com/lightning/bolts/issues/835
* https://github.com/lightningnetwork/lnd/pull/6703
* https://lists.linuxfoundation.org/pipermail/lightning-dev/2022-July/003643.html


joostjager:
I don't see how that achieves the same goal. I am looking to make senders aware of the discount so that they can choose routes that are better for them. How will a sender become aware of that new p2p message if it is only exchanged between two remote nodes?

joostjager:
Interesting idea. I suppose it's up to your counterparty to actually lower their outbound fee and not something that can be enforced. Which simplifies the implementation.

What's left is just the ability to tell your peer to send you fewer sats.

It may be a problem for final nodes though that also route. If they are accepting a payment as the final destination, they'll also receive less. The counterparty doesn't know which position in the route they are.

joostjager:
Well in the existing proposal, the sender is aware of the final hop. This BLIP says that no inbound fee needs to be paid to the final node, and the sender can take that into account when building the route.

Yes, a fake node can be used to let the payer pay for the discount they got :)

With the new p2p proposal, it may not be necessary anymore to do negative fees. That was just for backwards compatibility. I think it can be reformulated to positive fees. Basically forcing your peer to share part of their outbound fee with you.

For the final node, this would be a little better I guess because at least your getting the invoice amount.

joostjager:
I think that not having the ability to charge inbound fees is at this moment already a strong reason for node operators to reject / close certain inbound channels.

I can see this option becoming a requirement for those nodes. The win is that they can keep channels open that they would otherwise close.

blip-0018.md
```diff
@@ -0,0 +1,61 @@
...
+  These fields SHOULD be negative numbers that define the inbound discount for
+  that channel.
+  
+Receiving node:
+
+* MAY pay each routing node the sum of its advertised inbound and outbound fee
+for the incoming and outgoing channel that is used. This should amount to a
+lower value than just the outbound fee.
+
+  `outbound_fee(amt_to_fwd) + inbound_fee(amt_to_fwd + outbound_fee(amt_to_fwd))`
```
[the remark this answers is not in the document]
joostjager: Yes, considered that. But perhaps there is a future where all senders have upgraded and we shift to actual positive inbound fees, which would then be represented as negative discounts. Not optimally consistent. Of course this is all just cosmetics.

joostjager:
Need to think about what to do with the insufficient fee failure message. Maybe extend with the channel update message for the incoming channel?

joostjager:
Ah, this was a comment about the `channel_update` extension proposal. In the p2p solution, nothing like that is needed indeed.

One thing that worries me about the p2p proposal is that it probably requires changes to the channel state machine so that it will be clear from what point onwards exactly the new fee applies.

joostjager:
One downside of the p2p approach is that if the incoming node already charges zero fees, they can't lower their fee any further. So giving them a discount isn't going to make a difference. If there are multiple zero or low fee peers, it won't be possible to have distinct inbound fee schedules for these peers.

joostjager:
Yes, that's true. Over time, when senders are mostly upgraded, you could switch to positive inbound fees to get around the pathfinding limitation.

Although for p2p this may be uncomfortable for peers, because they need to monitor the inbound fee closely and adjust their fees accordingly to make sure they charge as much as they need to pay p2p to the next node and keep it net zero.

blip-0018.md
```diff
@@ -0,0 +1,92 @@
...
+  Senders typically require the inverse calculation to determine what the
+  `incoming_amt` should be. It's not completely straight-forward because of the
+  discontinuous integer division.
+
+  ```golang
+  if inbound_fee.proportional_millionths < 0 {
+    incoming_amt = divideRoundAwayFromZero((amt_to_receive+int64(inbound_fee.base_msat))*1e6, (1e6 - int64(inbound_fee.proportional_millionths)))
+  } else {
+    incoming_amt = divide((amt_to_receive+int64(inbound_fee.base_msat))*1e6, (1e6 - int64(inbound_fee.proportional_millionths)))
+  }
```
joostjager, line 77: Figured this out by trial-and-error and running a large numbers of values through the formula to see if it checks out. Would be nice if someone who really understands equations with integer math to validate.

joostjager:
I like the idea of the more intuitive fee calculation. Because we're talking about fees paid over fees, the difference is small. But it does simplify things on the implementation side. In particular the complicated inverse calculation that can now be dropped. Tested it, and it works without any unexpected challenges in pathfinding.

The other suggestion to not allow payment destinations to charge inbound fees is I think independent of this. For both ways of calculating, we can choose whether to make the final hop inbound-fee-free or not.

Charging inbound fees as a destination is already possible today by setting up a gateway node in front of the final destination. So in that regard there is not much fundamental change.

Perhaps this decision is better informed by looking at routing nodes that are also a payment destination. Suppose they charge inbound fees on some channels. Would it be desirable for them to charge the same fees if they receive a payment? I'd think that it is beneficial to be able to apply the same traffic shaping as for forwarded traffic, but interested to hear other's opinions.

joostjager:
>So I'd suggest calculating the fees over the outgoing amount only. By adding the the inbound + outbound fee policies together and applying that to the outgoing amount:

> total fee = 1,000,000 * (100ppm inbound fee + 200ppm outbound fee) = 300 sat

It's important to note that the policies first need to be added together. If applied separately, rounding differences may occur.

joostjager:
@JssDWt one potential downside of calculation inbound fees based on the downstream outgoing amount is that it may be difficult to use bolt11 route hints as is.

If the inbound fee was calculated over the upstream incoming amount, it may have been possible to fold the outbound and inbound fee for a particular channel into a single base fee and fee rate and communicate that to the sender.

joostjager:
@bitromortac suggested a third way to calculate the inbound fees: to base them on the outgoing htlc amount plus the outbound fee.

Example: inbound fee = 1 + 10%, outbound fee = 7 + 3%, outgoing htlc amount = 100

`---  122 ---> (in fee: 12) NODE (out fee: 10) --- 100 --->`

In this case, the amount that the inbound fee is based on is 100+10=110. The inbound fee is 1+10%(110)=12.

I've created a [Google Sheet](https://docs.google.com/spreadsheets/d/1eYIc4_i2LwpSB8-_nOGPp8ATh1w04NUY7LVMipXGCgw/edit?usp=sharing) that shows the three fee models side-by-side.

The advantage of this latest variation is that:
* No awkward inverse fee calculation required
* Fits better with pathfinding algorithms because the outgoing htlc amount doesn't need to be kept available.

joostjager:
Updated PR to reflect newest fee calculation described above.

halseth:
Regarding charging fees for incoming payments:
It seems like this is desirable to enable. If I'm a merchant, I could end up with some incoming channels being used more than others (perhaps because they are towards a well-connected routing node), so it would be nice to incentivise senders to use different channels by giving them a rebate.

blip-0018.md
```diff
@@ -0,0 +1,94 @@
...
+  underpayment of an invoice.
+
+* For `fee_insufficient` failures, the routing node SHOULD, in addition to the
+  outgoing channel policy, also return the incoming channel policy as a TLV
+  extension. But only if the sender signaled support for extended failure
+  messages.
+  
+  TBD: allocate type for this and describe.
+
+* MUST NOT set inbound fees on private channels. Inbound fees can't be
```
halseth, line 73: Is this true? Seem like it could be encoded in the invoice with a new feature bit, with similar backwards compatibility as forwarding payments (start with a discount, and senders that don't understand it will overpay).

joostjager: This should have been "can't _currently_ be communicated". We could add a new tag for the inbound fees, but would need to be careful not to duplicate too much information and increase the size of a qr code unnecessarily. 

Perhaps it's possible to do a TLV extension of the [`r` field](https://github.com/lightning/bolts/blob/16973e2b857e853308cafd59e42fa830d75b1642/11-payment-encoding.md?plain=1#L150), but not sure if decoders might trip over it.

I've also looked into folding the inbound and outbound fee into a single fee, but that doesn't seem to be possible. If for example the final destination requires an inbound fee, there is no way to express that through the outbound fee of the second-last node. It will still always only send the payment amount to the final hop and not more.

joostjager:
The main reason for dropping exit hops fees is that routing nodes that use inbound fees for forwards, can't turn these off if they are a payment destination. And exit hops fees can also be implemented using a virtual/phantom/lnmux channel.

I don't think this is compatible with "tell your peer to charge more", because they'd learn whether they are the second to last node.

joostjager:
For totally flexibility with regards to the exit hop fees, we can do a follow-up with a new optional `channel_update` tlv record that defines a base fee and rate specifically for senders to pay to the last hop.

joostjager:
>the restriction that the inbound fees are no longer allowed to be positive

Did I accidentally add this restriction somewhere? This is not the plan, the inbound fees can be either negative or positive. But in a transitionary period they will probably be negative.

I should definitely rename the blip though.

> If we go the "counterparty adds fee" route we don't have that limitation, and just end up with recipients charging fees. That's a bit awkward, sure, but recipients can already charge fees if they want to, so its not like its a new issue?

They can already do it currently if they want to, but the problem is that they can no longer not charge recipient fees if they don't want to. This is assuming that they do want inbound fees for regular forwards.

joostjager:
>For compatibility with the existing network, explicit inbound fees must always be negative. Senders should always ignore any positive inbound fees as nodes have to accept non-payment to remain compatible. I'm not sure what any sensible deployment strategy looks like here.

Start with negative inbound fees first to learn more about the practical utility of the concept. If a sufficiently large part of the senders takes inbound fees into account, routing nodes may move towards positive inbound fees and accept that they will lose out on routing fees for the remaining senders that ignore those fees.

I don't think it is useful for senders to always ignore positive inbound fees.

joostjager:
If a routing node chooses - at a later stage - to advertise and enforce a positive inbound fee, the sender should pay it. Otherwise their payment attempt will fail.

joostjager:
Senders would add support for that if they believe that inbound fees will make the network as a whole better. Maybe the BLIP can be promoted to a BOLT at that point. For now the initial goal is to experiment with negative fees and go from there.

joostjager:
>However, that means nodes can't use positive inbound fees until all senders have been upgraded, probably years, and we'd effectively "hard fork off" all unupgraded senders once that ships. 

I think it's fine if it takes years. WIth negative inbound fees and raised outbound fees, you can accomplish the exact same thing so there's no hurry. But if we eventually 'hard fork', we'll end up in a clean state. Similar to how we want to get rid of the legacy onion payload, even though there may be senders out there not supporting it.

joostjager:
The idea is that the negative fee provides an incentive for senders to upgrade. The negative fee and outbound raise can be tiny initially to not become unattractive compared to other routing nodes, but still enough to encourage upgrade.

And after senders upgrade, they'll run code that also deals with positive fees correctly if routing nodes at some point decide to require those.

What is the goal that you have in mind that a low/zero fee node would want to accomplish using inbound fees?

joostjager:
Another important point is the probability of my prs getting merged. I think that cutting deep into the channel state machine to change the balance update logic is not something that will be cheered for unanimously.

joostjager:
I think it is hard to predict how routing nodes are going to experiment with the new feature. Whether they do minimal negative fees, larger negative fees, apply positive fees to some channels only, etc. That's why I think it's interesting to allow this and observe what happens.

Regarding blip vs bolt: I just wanted to document the protocol change somewhere, but it starts to feel like asking for permission.

blip-0018.md
```diff
@@ -0,0 +1,94 @@
...
+    
+    `inbound fee = 1 + 10%, outbound fee = 7 + 3%, outgoing htlc amount = 100`
+
+    `--- 122 ---> (in fee: 12) (net_received_amt: 110) (out fee: 10) --- 100 --->`
+
+* MUST accept a routing fee equal to or greater than the sum of inbound and
+  outbound fee. No attempt should be made to calculate the total fee in one go,
+  because rounding may cause the end result to be higher.
+
+* Final hops MUST accept a routing fee of at least:
```
joostjager, line 59: Need to remove exit hop fees from blip.

JssDWt:
> Why would any sender add support for that - if no one supports it no routing node will ever turn it on.

@TheBlueMatt Senders will start supporting inbound fees because it may make payments cheaper for them at first, because routing nodes charge negative inbound fees at first. Why would a sender implement this blip only for negative inbound fees? That doesn't make sense to me. There's only a few lightning implementations, and they would implement inbound fees in both the negative and positive case for the sender, because if they wouldn't implement the positive fee case, the sender would get errors, which would hurt the UX.

blip-0018.md
```diff
@@ -0,0 +1,90 @@
...
+    
+    `inbound fee = 1 + 10%, outbound fee = 7 + 3%, outgoing htlc amount = 100`
+
+    `--- 122 ---> (in fee: 12) (net_received_amt: 110) (out fee: 10) --- 100 --->`
+
+* MUST accept a routing fee equal to or greater than the sum of inbound and
+  outbound fee. No attempt should be made to calculate the total fee in one go,
+  because rounding may cause the end result to be higher.
+
+* The inbound fee does not apply to final hops.
```
thomash-acinq, line 59: I think this shows what's wrong with this proposal : If I open a channel to X, I should be able to pay X without paying fees, it would be unfair for X to ask me to pay an extra fee for receiving my payment. Well, it's exactly the same in case X is not the final recipient but only a relay. If X really thinks that traffic from channel a is better than traffic from channel b, then why should this preference no longer hold when X is the recipient?

joostjager: Because the recipient isn't providing a forwarding service. It plays a different role.

thomash-acinq: Why does it matter? Traffic is fungible.

joostjager: I'd say it is different traffic, because there is no outgoing component. There is something to say for always charging, but I think it is not the best choice. I documented some thoughts about it [here](https://github.com/lightning/blips/pull/18#issuecomment-1284437820)

blip-0002.md
```diff
@@ -97,6 +97,14 @@ The following table contains extension tlv fields for the `ping` message:
 |-------|-----------------------------|--------------------------------|
 | 65536 | `tlv_field_name`            | Link to the corresponding bLIP |
 
+#### `channel_update`
+
+The following table contains extension tlv fields for the `channel_update` message:
+
+| Type       | Name                        | Link                           |
+|------------|-----------------------------|--------------------------------|
+| 55555      | `inbound_discount`          | [bLIP 18](./blip-0018.md)      |
```
feelancer21, line 106: The current implementation in lnd also applies to optional positive inbound fees. Should the wording be updated?

blip-0018.md
```diff
@@ -0,0 +1,90 @@
...
+
+    Example: 
+    
+    `inbound fee = 1 + 10%, outbound fee = 7 + 3%, outgoing htlc amount = 100`
+
+    `--- 122 ---> (in fee: 12) (net_received_amt: 110) (out fee: 10) --- 100 --->`
+
+* MUST accept a routing fee equal to or greater than the sum of inbound and
+  outbound fee. No attempt should be made to calculate the total fee in one go,
+  because rounding may cause the end result to be higher.
```
feelancer21, line 57: In the current lnd implementation, the sum of outbound and inbound must not be negative. Should the blip be updated accordingly? 

I think that negative fees should also be possible in the long term, but then we would probably need an additional boolean to signal a possible negative fee to the network. Otherwise sender nodes will not be able to see which routing might reject a payment with negative fees.

PR #23, opened 2023-04-20
Bip 0010 cleanup

satoshisstream:
Removing unused properties, clearing up fields, adding the `auto` option.

More info: https://github.com/Podcastindex-org/podcast-namespace/pull/510


Roasbeef (approved):
LGTM 🐲

PR #24, opened 2023-04-25
blip2: add advisory for invoice feature bit maximum

carlaKC:
Tagged fields in bolt11 invoices are limited to 10 bits to express the length of the field, which means we have 1023 * 5 ( = 5115) bits for features. Including a 0 position feature, this means that invoice feature bits must be <= 5114. In practice we hope and pray to never see 639 byte feature vectors, this is worth accounting for when implementing custom features.

This limitation does not apply in other contexts, because feature vectors are encoded with (u16) variable length.


dunxen:
Concept ACK

t-bast (approved):
LGTM, this is a small clarification on an existing blip, merging this now.

PR #26, opened 2023-06-07
blip-0026: L402 - Lightning HTTP 402 Protocol 

Roasbeef:
In this commit, we add a bLIP for the L402 (formerly known as LSAT)
protocol. The L402 protocol presents a standardized way of adding LN
micropayments to any existing HTTP-REST or gRPC API. The L402 protocol
repurposes the HTTP 402 Payment Required error code with the necessary
authentication headers required to bind a request's validity to the
payment of an LN invoice. Macaroons are used as flexible authentication
credentials (supports custom caveats, attenuation, etc) which allow an
L402 reverse-proxy to validate an L402 API key without backend LN node
interaction for each request (the macaroon commits to the payment hash).


Roasbeef:
cc @wpaulino @guggero @kodylow

TheBlueMatt (commented):
Awesome! Thanks for writing this up, this is nifty. The auth protocol described here is wonderfully simple and deserves to be specified, but this document is hugely verbose and adds a wall of text and discussion that makes this protocol look downright complicated and scary. I'm confident this could be cut down by at least half without any less ability to communicate all its detail and make it much less scary to a dev who just wants to use it :)

blip-0026.md
```diff
@@ -0,0 +1,712 @@
...
+2. As long as the credential has not expired, the steps 9-13 above will be
+   followed.  If/when the credential expires, the server will start over at
+   step 4 and instruct
+   the client to obtain a fresh credential.
+
+
+## L402 HTTP/gRPC Protocol Specification
+
+This section defines the "L402" authentication scheme, which transmits
+credentials as `<macaroon(s)>:<preimage>` pairs, where the preimage is encoded
```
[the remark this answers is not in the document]
dstadulis: The cryptographic properties of macaroons are desired, so the distinction is intentional.
Resources expounding on the application utility of these cryptographic properties are here: https://docs.lightning.engineering/the-lightning-network/l402/macaroons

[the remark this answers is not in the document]
Roasbeef: >  and the thing that the server returns may be a macaroon, but from the perspective of this protocol, its just an opaque token.

The requirement here is that the API key itself commits to the payment hash within the invoice. This attribute is central to the protocol. Committing to the payment hash in the api key is what allows the reverse proxy implementing this protocol to be more or less stateless: given the pre-image and the macaroon (symmetric key hmac chain that includes a commitment to the payment hash in the chain), it can validate if an API key is complete or not. The API key here is a two-tuple of `(macaroon, preimage)`. 

This structure means that the reverse proxy doesn't need to maintain all the invoice state of the LN node it's connected to, and also doesn't need to hit an API to see if an invoice has been paid or not for each request. At high loads, such a requirement would prohibit many use cases. 

It's feasible that one could use JWT's or w/e (inheriting all the baggage) other structured API keys here, but then we'd also have to specify how for each of those, the committed payment hash is to be parsed and recognized. Macaroons are more or less less standardized in their structure (tons of implementations in various languages that can all parse the same macaroons), and in use elsewhere in other independent ecosystems. 

Devs are totally free to spin a similar protocol using PASETO or w/e another JWT alternatives of course. The goal of this document is to specify the L402 protocol as already defined and deployed in the wild. It's also worth noting that these headers are just for the L402 portion, the server being proxied to gets these headers and can interpret them if they wish, but can also layer w/e other authentication mechanism (oauth, etc, etc) as well.

blip-0026.md
```diff
@@ -0,0 +1,712 @@
...
+example is automated tier upgrades. Many APIs typically offer several tiers
+which allow users to gain access to more or additional resources as they climb
+up the ladder. Typically, a user must _manually_ navigate a web-page to request
+an upgrade to a higher tier, or downgrade to a lower tier. With the L402
+standard, tier upgrades can easily be automated: the user hits a new endpoint
+to obtain an _upgraded_ L402 which _encodes_ additional functionality or
+increased resource access compared to the prior tier. Services can even
+leverage L402 for A/B Testing by giving subsets of users a distinct L402 which
+when submitted to the service, render a slightly different version of the
+target resource or service.
```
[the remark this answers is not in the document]
studioTeaTwo: I think it's helpful to have a historical story to grab the "background" quickly.

Though, considering that this is a bLIP, certainly, I also feel that the subsequent parts lack the work to abstract, generalize and standardize.

This conception is great. I hope a future where this is seen in IETF/W3C documents.

niftynei:
hey yall, i've been making some L402 libraries for the CLN ecosystem. Here's a bit of feedback.

- This doc contains way more than necessary to implement the L402 protocol. Sometimes more verbiage + info can be useful, but in this case it is actually a disservice towards helping people build or re-implement the spec for themselves. I say this having had to parse it to write a conforming Flask interceptor.
- the usage of the term `macaroons` in the spec is unnecessarily restrictive. you can use any token/hmac system that you'd like. I've been using `token` in my implementation.

Cool protocol, really nice that this finally is getting some proper documentation that we can do cross implementation on

blip-0026.md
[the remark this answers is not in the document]
niftynei: Kinda feel like there's a disagreement here in terms of what's in scope of the L402 nomenclature.

I'd classify it as the "data pieces and their format in the HTTP headers (or grpc calls)". Feels like there's a strong opinion herein to also include the authentication token format and specification, but it's pretty obvious to me that these are separable/unimportant from a *client's* perspective.

For any client that wants to know how to parse and pay/return a call from an L402 response, the auth token format is a distraction. In that case, the "data pieces and format" are all that really matter.

If there's other data or formatting that we should consider, that's probably worth documenting here. But the server side security/auth and/or how stateful/stateless it needs to be seems like an implementation detail.

> This structure means that the reverse proxy doesn't need to maintain all the invoice state of the LN node it's connected to, and also doesn't need to hit an API to see if an invoice has been paid or not for each request. 

This is a nice property that you can implement in any way/with a variety of tokens, but I'd argue that it's a "nice to have" of what you *can* do with the L402 tokens+preimages. You can also dumbly store state and do lookups (which you kinda have to do anyway for time-based access tokens / and or there different approaches to this).

blip-0026.md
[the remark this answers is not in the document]
Roasbeef: Thx for the feedback, I've condensed much of this in the latest iteration. Likely more can be cut in future iterations.

blip-0026.md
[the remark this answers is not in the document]
Roasbeef: >  but it's pretty obvious to me that these are separable/unimportant from a client's perspective.

Agreed, note that this information is only really relevant to the L402 proxy. The web service may still require the client to attach normal authentication information such as cookies or w/e other custom information. In most cases, the L402 proxy can be oblivious about the actual request information it's proxying, it only needs to ensure all the headers are copied over as normal. A client also doesn't have to introspect into the credential at all, it just encodes it with base64, then puts it where the server expects it. 

The expectation isn't that all active web services switch over to macaroons, it's that only the proxy needs to care about it, and otherwise the API backend can go unchanged. 

I've updated these sections to mention the optimization (committing to the hash in the credential), and weakened the language around verification.

Roasbeef:
Thanks for the feedback @niftynei!

> This doc contains way more than necessary to implement the L402 protocol

The latest push significantly cuts down on the size of the document. Further edits are likely possible as well near the start and end of the specification. 

> the usage of the term macaroons in the spec is unnecessarily restrictive

The latest iteration weakens the language around macaroons, and moves the section on a suggested structure to the appendix. You're correct that anything can be used in practice, IMO given the macaroons are already specified and standardized, I think it makes sense to nudge proxy implementations in that direction. Note that clients don't need to introspect into them at all, and just serialize them where expected. It also isn't expected the API backends rely soley on L402 for in-depth authentication, normal auth headers or cookies or w/e can still be used. The proxies are meant to be loosely coupled to the backends they support, such that any API can be put in front of a proxy and Things Just Work^TM.

blip-0026.md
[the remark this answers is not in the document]
niftynei: proxy // in-app HTTP handler middleware is baaaaaasically the same thing only difference is which machine it's running on, no?

I don't understand how *where* the L402 is being handled has a material impact on the specification of the "data pieces and their format" which is kiinda all we need from a spec?

blip-0026.md
```diff
@@ -0,0 +1,641 @@
...
+    S->>C: HTTP/1.1 200 OK
+```
+
+HTTP-based protocol flow:
+
+1. A client hits an `HTTP` endpoint to fetch a resource.
+
+2. The server decides that payment is required to access the endpoint. They
+   return a HTTP status code of `402` along with a `WWW-Authenticate` header
+   containing a macaroon and invoice. The macaroon commits to the payment hash
```
niftynei, line 177:
```suggestion
   containing a token and invoice. Ideally, the token commits to the payment hash
```

blip-0026.md
```diff
@@ -0,0 +1,641 @@
...
+   of the invoice `H`.
+
+3. The client pays the invoice over the Lightning Network and receives the
+   payment pre-image `r` to the payment hash `H`, `H = sha256(r)`
+
+4. The client re-issues the request with a `Authorization` header that includes
+   the payment preimage `r`, and the macaroon sent by the server.
```
niftynei, line 184:
```suggestion
   the payment preimage `r`, and the token sent by the server.
```

blip-0026.md
```diff
@@ -0,0 +1,641 @@
...
+
+5. The server verifies the cryptographically verifies the authenticity and
+   integrity of the macaroon, then ensures the payment hash committed to in the
```
niftynei, line 187:
```suggestion
   integrity of the token, then ensures the payment hash committed to in the
```

blip-0026.md
```diff
@@ -0,0 +1,641 @@
...
+   macaroon `H`, satisfies the relation the `H = sha256(r)`.
```
niftynei, line 188:
```suggestion
   token `H`, satisfies the relation the `H = sha256(r)`.
```

blip-0026.md
```diff
@@ -0,0 +1,641 @@
...
+
+6. The server returns a 200 OK error code along with the specified resource.
+
+If the service supports persistent L402 credential re-use, then a client SHOULD
```
niftynei, line 192:
```suggestion
If the service supports persistent L402 token re-use, then a client SHOULD
```

?? not sure about this one. Do you mean something other than the auth/hmac token here?

blip-0026.md
```diff
@@ -0,0 +1,641 @@
...
+new challenge in response to a client request containing a newly invalid L402.
+Possible revocation conditions include: expiry date, exceeded N usages, volume
+of usages in a certain time period necessitating a tier upgrade, and
+potentially others \(discussed further in the higher-level design document\).
+
+L402 could be configured for use on a per-backend-service basis or for all
+exposed services. As an example, the same L402 credential may commit to
+capabilities enabling a user to access both service `Y` and `X from the same
+endpoint. This flexibility is afforded because all services are going to be
+gated by the same L402 proxy, which verifies all macaroons for all backend
```
niftynei, line 210:
```suggestion
gated by the same L402 proxy, which verifies all credentials for all backend
```

blip-0026.md
```diff
@@ -0,0 +1,641 @@
...
+* `Y`: is the Lightning Network invoice a client MUST pay in order to obtain
+  a pre-image required to complete the `Authenticate` header.
+
+* `X`: is the macaroon credential that encodes the client's capabilities, and
+  also cryptographically commits to the payment hash `H` of the invoice `Y`.
+
+### Client HTTP Authentication Headers
+
+This section defines the client-side L402 authentication headers, which
+transmits credentials as `<macaroon(s)>:<preimage>` pairs, where the preimage
```
niftynei, line 255:
```suggestion
transmits credentials as `<token(s)>:<preimage>` pairs, where the preimage
```

blip-0026.md
```diff
@@ -0,0 +1,641 @@
...
+is encoded as hex and the macaroon is encoded as base64. Multiple macaroons are
```
niftynei, line 256:
```suggestion
is encoded as hex and the token is either encoded as base64 or hex. Multiple tokens are
```

blip-0026.md
```diff
@@ -0,0 +1,641 @@
...
+base64 encoded individually and listed comma separated before the colon.
```
niftynei, line 257:
```suggestion
encoded individually and listed comma separated before the colon.  
```

blip-0026.md
```diff
@@ -0,0 +1,641 @@
...
+
+### Client HTTP Authentication Headers
+
+This section defines the client-side L402 authentication headers, which
+transmits credentials as `<macaroon(s)>:<preimage>` pairs, where the preimage
+is encoded as hex and the macaroon is encoded as base64. Multiple macaroons are
+base64 encoded individually and listed comma separated before the colon.  
+
+The L402 authentication scheme is based on the model that the client needs to
+authenticate itself with a macaroon and invoice preimage for each backend
```
niftynei, line 260:
```suggestion
authenticate itself with a token and invoice preimage for each backend
```

blip-0026.md
```diff
@@ -0,0 +1,641 @@
...
+service it wants to access. The server will service the request only if it can
+validate the macaroon and preimage for the particular backend service
```
niftynei, line 262:
```suggestion
validate the access token and preimage for the particular backend service
```

blip-0026.md
```diff
@@ -0,0 +1,641 @@
...
+requested.
+
+The L402 authentication scheme utilizes the Authentication Framework specified
+in [_RFC 7235_](https://tools.ietf.org/html/rfc7235) as follows.
+
+In challenges: the scheme name is "L402". Note that the scheme name is
+case-insensitive. For credentials, the syntax is:
+
+`macaroons` → [_&lt;base64 encoding&gt;_](https://tools.ietf.org/html/rfc3548#section-3), comma separated if multiple macaroons are present
```
niftynei, line 271:
```suggestion
`auth tokens` → [_&lt;hex | base64 encoding&gt;_](https://tools.ietf.org/html/rfc3548#section-3), comma separated if multiple auth tokens are present
```

blip-0026.md
```diff
@@ -0,0 +1,641 @@
...
+`preimage` → [_&lt;hex encoding&gt;_](https://tools.ietf.org/html/rfc3548#section-6)
+`token` → macaroons ":" preimage
```
niftynei, line 273:
```suggestion
`L402 credential` → auth tokens ":" preimage
```

blip-0026.md
```diff
@@ -0,0 +1,641 @@
...
+
+Specifically, the syntax for "token" specified above is used, which can be
```
niftynei, line 275:
```suggestion
Specifically, the syntax for "L402 credential" specified above is used, which can be
```

blip-0026.md
```diff
@@ -0,0 +1,641 @@
...
+considered comparable to the [_"token68"
+syntax_](https://tools.ietf.org/html/rfc7235#section-2.1) used for HTTP basic
+auth.
+
+A valid client authorization header takes the form:
+```
+Authorization: L402 macaroon:preimage
```
niftynei, line 282:
```suggestion
Authorization: L402 token:preimage
```

blip-0026.md
```diff
@@ -0,0 +1,641 @@
...
+
+Upon receipt of a request for a URI of an L402-proxied backend service that
+lacks credentials or contains an L402 that is invalid or insufficient in some
+way:
+
+1. Ths server SHOULD derive a economical price for the endpoint expressed in
+   milli-satoshis (1/1000th of a satoshi) and create a BOLT 11 invoice `P`
+   requesting that amount to its backing Lightning Node.
+
+2. The server MUST create a new authentiation credential `M` for the client
```
niftynei, line 300:
```suggestion
2. The server MUST create a new auth token `T` for the client
```

blip-0026.md
```diff
@@ -0,0 +1,641 @@
...
+   that will only become valid once the client pays the LN invoice.
+
+    3.1. The server SHOULD ensure that the authentiation credential commits to
```
niftynei, line 303:
```suggestion
    3.1. The server SHOULD ensure that the auth token commits to
```

blip-0026.md
```diff
@@ -0,0 +1,641 @@
...
+
+4.  Alongside the 402 status code, the server MUST should specify the
+    `WWW-Authenticate` header \([_\[RFC7235\], Section
+    4.1_](https://tools.ietf.org/html/rfc7235#section-4.1)\) field to indicate
+    the L402 authentication scheme and the macaroon needed for the client to
+    form a complete L402.
+
+    A valid response header MUST take the form of:
+    ```
+    WWW-Authenticate: L402 macaroon="M", invoice="P"
```
niftynei, line 323:
```suggestion
    WWW-Authenticate: L402 token="T", invoice="P"
```

blip-0026.md
```diff
@@ -0,0 +1,641 @@
...
+    ```
+
+    An example L402 HTTP response resembles: 
+
+    ```text
+    HTTP/1.1 402 Payment Required
+
+    Date: Mon, 04 Feb 2014 16:50:53 GMT
+
+    WWW-Authenticate: L402 macaroon="AGIAJEemVQUTEyNCR0exk7ek90Cg==", invoice="lnbc1500n1pw5kjhmpp5fu6xhthlt2vucmzkx6c7wtlh2r625r30cyjsfqhu8rsx4xpz5lwqdpa2fjkzep6yptksct5yp5hxgrrv96hx6twvusycn3qv9jx7ur5d9hkugr5dusx6cqzpgxqr23s79ruapxc4j5uskt4htly2salw4drq979d7rcela9wz02elhypmdzmzlnxuknpgfyfm86pntt8vvkvffma5qc9n50h4mvqhngadqy3ngqjcym5a"
```
niftynei, line 333:
```suggestion
    WWW-Authenticate: L402 token="AGIAJEemVQUTEyNCR0exk7ek90Cg==", invoice="lnbc1500n1pw5kjhmpp5fu6xhthlt2vucmzkx6c7wtlh2r625r30cyjsfqhu8rsx4xpz5lwqdpa2fjkzep6yptksct5yp5hxgrrv96hx6twvusycn3qv9jx7ur5d9hkugr5dusx6cqzpgxqr23s79ruapxc4j5uskt4htly2salw4drq979d7rcela9wz02elhypmdzmzlnxuknpgfyfm86pntt8vvkvffma5qc9n50h4mvqhngadqy3ngqjcym5a"
```

blip-0026.md
```diff
@@ -0,0 +1,641 @@
...
+    ```
+
+    where `"AGIAJEemVQUTEyNCR0exk7ek90Cg=="` is the macaroon that the client
```
niftynei, line 336:
```suggestion
    where `"AGIAJEemVQUTEyNCR0exk7ek90Cg=="` is the auth token that the client
```

blip-0026.md
```diff
@@ -0,0 +1,641 @@
...
+    preimage that must be included for each of its authorized requests.
+
+Upon receiving an HTTP request with a valid RFC-7235 `Authorization` header as
+defined in the Client HTTP Authentication Headers section:
+
+1. The server MUST verify the cryptographic authenticity and integrity of the
+   credential.
+  1.1. If a credential is invalid, then the server MUST return a 401
+  Unauthorized status code.
+2. The server MUST parse out the `Authorization` header (`macaroon:preimage`)
```
niftynei, line 348:
```suggestion
2. The server MUST parse out the `Authorization` header (`auth token:preimage`)
```

blip-0026.md
```diff
@@ -0,0 +1,641 @@
...
+   into a `M` the encoded credential, and `r` the payment preimage.
```
niftynei, line 349:
```suggestion
   into a `T` the encoded auth token, and `r` the payment preimage.
```

blip-0026.md
```diff
@@ -0,0 +1,641 @@
...
+3. The server MUST verify that the BOLT 11 invoice tied to the credential `M`
```
niftynei, line 350:
```suggestion
3. The server MUST verify that the BOLT 11 invoice tied to the auth token `T`
```

blip-0026.md
```diff
@@ -0,0 +1,641 @@
...
+   has been paid.
+   3.1. If the server committed to the payment hash `H` in the credential `M`,
```
niftynei, line 352:
```suggestion
   3.1. If the server committed to the payment hash `H` in the auth token `T`,
```

blip-0026.md
```diff
@@ -0,0 +1,641 @@
...
+Upon receiving a valid L402 `WWW-Authorization` header the client:
+
+1. Should verify that the contained BOLT 11 invoice doesn't request an
+   excessive amount of Bitcoin.
+1.1. If so, then the client SHOULD abandon the HTTP request.
+2. After validating the returned BOLT 11 invoice, the client MUST pay the
+   invoice over the LN to receive the payment pre-image `r`.
+3. The client then constructs a valid `Authentication` header with the
+   following form:
+   `Authorization: L402 macaroon:preimage`
```
niftynei, line 378:
```suggestion
   `Authorization: L402 auth_token:preimage`
```

blip-0026.md
```diff
@@ -0,0 +1,641 @@
...
+4. The client then re-issues the original HTTP request with the attached L402
+   `Authorization` header.
+
+In other words, to receive authorization, the client:
+
+* Pays the invoice from the server, thus revealing the invoice’s preimage
+* Constructs the L402 by concatenating the base64-encoded macaroon\(s\), a
```
niftynei, line 385:
```suggestion
* Constructs the L402 by concatenating the encoded auth token\(s\), a
```

blip-0026.md
```diff
@@ -0,0 +1,641 @@
...
+  single colon \(":"\), and the hex-encoded preimage.
+
+Since the macaroon and the preimage are both binary data encoded in an ASCII
```
niftynei, line 388:
```suggestion
Since the auth token and the preimage are both binary data encoded in an ASCII
```

blip-0026.md
```diff
@@ -0,0 +1,641 @@
...
+based format, there should be no problem with either containing control
+characters or colons \(see "CTL" in [_Appendix B.1 of
+\[RFC5234\]_](https://tools.ietf.org/html/rfc5234#appendix-B.1)\). If a user
+provides a macaroon or preimage containing any of these characters, this is to
```
niftynei, line 392:
```suggestion
provides an auth token or preimage containing any of these characters, this is to
```

blip-0026.md
```diff
@@ -0,0 +1,641 @@
...
+be considered an invalid L402 and SHOULD result in a 402 and authentication
+information as specified above.
+
+If a client wishes to send the macaroon `"AGIAJEemVQUTEyNCR0exk7ek90Cg=="`
```
niftynei, line 396:
```suggestion
If a client wishes to send the auth token `"AGIAJEemVQUTEyNCR0exk7ek90Cg=="`
```

blip-0026.md
```diff
@@ -0,0 +1,641 @@
...
+the payment hash of an invoice, this is a trivial step.
+
+The final step ensures a macaroon is authorized to access a service. This is
+done by ensuring the service-specific caveat predicates of a macaroon hold true
+for the service being accessed. If only one of these caveats doesn’t hold true,
+then the macaroon is invalid. In the case of an unknown caveat, its evaluation
+must be skipped by the authorizer as the macaroon holder can further attenuate
+the macaroon for other applications.
+
+### Macaroon Revocation
```
niftynei, line 623:
```suggestion
### Auth Token Revocation
```

niftynei (commented):
Added suggested comments to update (at least the HTTP section) to conform to the nomenclautre i'm using in `chargesats` (macroon -> token). 

Moving the Specification up within the first 200 lines is a massive improvement from befrore. :ok_hand:

blip-0026.md
```diff
@@ -0,0 +1,641 @@
...
+standard supports _gRPC_ which supports _bi directional streaming_ APIs, one
+could even create a metered streaming video or audio service as well!
+
+Additionally, L402 also enables innovation at the API _architecture_ level. One
+example is automated tier upgrades. Many APIs typically offer several tiers
+which allow users to gain access to more or additional resources as they climb
+up the ladder. Typically, a user must _manually_ navigate a web-page to request
+an upgrade to a higher tier, or downgrade to a lower tier. With the L402
+standard, tier upgrades can easily be automated: the user hits a new endpoint
+to obtain an _upgraded_ L402 which _encodes_ additional functionality or
```
bucko13, line 138: Just noting this part here wrt to the conversations about if macaroons are implementation detail or part of the spec. Using macaroons for the token is part of what makes this possible since [attenuation](https://hackingdistributed.com/2014/05/16/macaroons-are-better-than-cookies/) like this is part of the design goal of macaroons.

blip-0026.md
```diff
@@ -0,0 +1,641 @@
...
+In this section, we specify the protocol for the HTTP portion of the L402
+proxy.
+
+#### Server Flow
+
+Upon receipt of a request for a URI of an L402-proxied backend service that
+lacks credentials or contains an L402 that is invalid or insufficient in some
+way:
+
+1. Ths server SHOULD derive a economical price for the endpoint expressed in
```
bucko13, line 296:
```suggestion
1. Ths server SHOULD derive an economical price for the endpoint expressed in
```

blip-0026.md
```diff
@@ -0,0 +1,641 @@
...
+specification_](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md).
+
+# Universality
+
+# Backwards Compatibility
+
+The `L402` protocol was formerly known as `LSAT`. In order to preserve
+backwards compatibility with clients, anywhere `L402` is used within HTTP
+headers or requests, a valid protocol flow with the string `LSAT` should also
+be accepted.
```
bucko13, line 455: 🎉

blip-0026.md
```diff
@@ -0,0 +1,641 @@
...
+Lightning API key \(L402\). To recap, an L402 is composed of a macaroon, which
+specifies the allowed capabilities of the credential, and a preimage, which
+serves as the credential’s proof of payment. Macaroons are perfect candidates
+for L402 as they are tamper-proof, support easy rotation, have attributes, and
+can even be further attenuated in order for applications that integrate an L402
+enabled service to delegate any additional capabilities. We’ll cover how
+macaroons are created, attenuated, and verified as part of L402. This chapter
+will require an understanding of how macaroons work and how they are useful in
+the context of authentication. It may be useful to skim the [introductory
+research paper on macaroons](https://research.google/pubs/pub41892/).
```
bucko13, line 471: Might be good to reference some implementations as well. Not all of them are interoperable unfortunately, so highlighting ones that are standard compliant would be useful as well as for further reference on how they work which can be helpful to understanding.

blip-0026.md
```diff
@@ -0,0 +1,641 @@
...
+for the service being accessed. If only one of these caveats doesn’t hold true,
+then the macaroon is invalid. In the case of an unknown caveat, its evaluation
+must be skipped by the authorizer as the macaroon holder can further attenuate
+the macaroon for other applications.
+
+### Macaroon Revocation
+
+To prevent abusers of a macaroon-based authenticated service, a macaroon should
+be able to be revoked. This can be achieved by having the minter remove the
+macaroon’s corresponding root key. By doing so, the minter will never be able
```
bucko13, line 627: Unfortunately this makes the system somewhat less stateless :-/

bucko13 (approved):
Woot! Exciting to see this being formalized and to hear other people are working on it 🎉

blip-0026.md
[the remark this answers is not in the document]
bucko13: So, I can see why there might be an argument to keeping the specification here agnostic on the encoding of the token itself, but as someone who has implemented this as a [middleware already](https://github.com/tierion/boltwall), I do think there are benefits to having this specified as well. By standardizing the full encoding you make the L402 token itself much more portable such that clients that don't have to have a relationship with the server itself in order to know how to decode and validate the macaroon. If the encoding is standardized, then any consumer of the L402 will know how to deserialize the macaroon portion, read the payment hash and validate against the preimage. 

There's also ecosystem benefits to having a system like this be cross-compatible. As an example, using the [LSAT Playground](https://lsat-playground.vercel.app/) (I guess I need to rebrand this now...) you can validate an L402 generated from boltwall or aperture, you can even take an L402 generated from aperture, [add a caveat](https://lsat-playground.vercel.app/#caveats) in the playgrond, and then have it validate against a server running boltwall middleware. You get additional benefits imo when you want to start running in environments like Nostr (if we could get L402 auth [standardized](https://github.com/nostr-protocol/nips/compare/master...bucko13:nips:402-auth) for nostr) where you might have different clients and servers needing to interoperate, or something like https://sphinx.chat. 

This is particularly useful when caveats come in to play so that clients can introspect their token to see how much time might be left before their authorization expires for example. So all of this plays a pretty important role in the nature of these being stateless: even though it could be encoded in other ways, having it be standardized makes it far more portable. 



> I don't understand how where the L402 is being handled has a material impact on the specification of the "data pieces and their format" which is kiinda all we need from a spec?

Aren't the "data pieces and their format" covered by specifying the token being a base64 encoded macaroon?

TheBlueMatt (commented):
Thanks for making this more readable. I think still a good chunk of text could be cut down.

slavakurilyak:
Any updates on this since last year?

/cc @roasbeef

ProofOfKeags:
> Any updates on this since last year?

What updates are you looking for specifically?

blip-0026.md
Roasbeef: It's purely optional, otherwise you can implement time based revocation just via a caveat expiry. I think it also depends on the service, some may be able to get away with a single global service key, while others may already require per user/account keys for the macaroons. In theory, you can have the best of both worlds if a deterministic scheme is used to derivation of the symmetric keys.

Roasbeef:
@slavakurilyak Hi, I've been gathering some significant feedback from a dozen or so teams/companies actively working on integrating the protocol into their offerings. We've been meeting once a month or so to discuss improvements to the protocol and also generally how things are fairing in the wild. I plan to make future PRs to start to incorporate some of this feedback (eg: ways to handle more stateful APIs, structured information outside the macaroon caveats, etc, etc). 

With all that said, I'll move to merge this in a few days (addressing the obvious typos/wording improvements suggested above), as the current text has been adequate for those seeking to actively build upon the protocol, and some felt that the lack of a merged PR meant the protocol was in an ambiguous state.

blip-0026.md
```diff
@@ -0,0 +1,641 @@
...
+hash+preimage relation.
+
+
+#### Client Flow
+
+Upon receiving a valid L402 `WWW-Authorization` header the client:
+
+1. Should verify that the contained BOLT 11 invoice doesn't request an
+   excessive amount of Bitcoin.
+1.1. If so, then the client SHOULD abandon the HTTP request.
```
positiveblue, line 373: format: same here

positiveblue (approved):
Thank you for the push @Roasbeef, I'm thrilled about this! 🚀

Regarding L402, I would suggest focusing on macaroons and having the spec define how caveats are structured (as we do here). Later, we can even include n appendix on "standard" caveats (e.g., expires vs expires_at). But I would do that once we observe their usage in the wild.

I'd like to propose two additions:

**Versioning in the `WWW-Authenticate` header**: Adding a version field apart from "macaroon" and "invoice" would allow for different "encoding" in the future, enabling clients/servers to understand and respond appropriately. If we call the current version `V0`

```
 L402 version="V0" macaroon="X", invoice="Y"
 ```


**Returning multiple macaroons/invoices**: While the exact implementation in the WWW-Authenticate header is uncertain, it would be beneficial to return an array of "Offers" with specific macaroons/invoices for each one, along with data explaining their purpose (like "supported plans" and what each plan includes). If versioning is part of the protocol (🤞 ), we can fix this in the future and even return them in the response body instead of the header in future versions of the protocol.

File: blip-0026.md [hunk 39]
```diff
@@ -0,0 +1,641 @@
...
+signature. Minting a new macaroon only requires the public identifier and the
+root key it corresponds to.
+
+Each macaroon must have its own cryptographically random root key that must not
+be disclosed to prevent forgeability of macaroons. Each root key consists of 32
+bytes, which provides a reasonable tradeoff between entropy and size. Root keys
+are essential to the verification of macaroons, so they must be stored securely
+and reliably.
+
+#### Macaroon Identifier
```
positiveblue, line 486: I think we may be over specifying a bit here.

Why wouldn't `user_id`/`macaroon_version` be caveats? 

An L402 credentials are linked to a specific challenge (invoice) and what identifies that invoice is the `payment_hash` so I would go with that in my implementation and leave everything else to caveats.

Not sure why this needs to be specified though? Unlike [biscuits](https://www.biscuitsec.org), macaroons can only be validated by the issuer, so the id should be whatever the issuer wants without encoding extra information/encoding whatever the issuer wants.

blip-0026.md
```diff
@@ -0,0 +1,641 @@
...
+In order to signal that L402 authentication is required, the server sends a
+[RFC-7235](https://datatracker.ietf.org/doc/html/rfc7235#section-4.1)
+authentication challenge. A challenge takes the form of:
+```
+WWW-Authenticate: challenge
+```
+
+For the L402 protocol the challenge takes the following form:
+```
+L402 macaroon="X", invoice="Y"
```
positiveblue, line 241: Are the `"` needed? 

[Aperture](https://github.com/lightninglabs/aperture/blob/master/l402/header.go#L32), [Fewsats](https://github.com/Fewsats/fewsatscli/blob/main/credentials/credentials.go#L77) and  [lsat-js](https://github.com/Tierion/lsat-js/blob/master/src/lsat.ts#L20) do not add them but [Sulu](https://github.com/sulusolutions/l402-ts/blob/main/src/client/axios.ts#L68) does. I think it is because of how it is specified here. 

If I could choose, I would say let's go without them.

SamuelAl: I think you are referring to two different things here @positiveblue . From what I see Aperture does add the `"` double-quotes when generating the "challenge" header, as can be seen [here](https://github.com/lightninglabs/aperture/blob/3a8936c4480425c0dd40a4b97864510a0fd78cc4/auth/authenticator.go#L110). Sulu replicates this functionality, as does [lsat-js](https://github.com/Tierion/lsat-js/blob/d902b9553bfb371f3c74773a80d65d3b35c50a81/src/lsat.ts#L30C3-L33C4). 

It is the challenge response with the L402 token that does not use `"`  , as can be seen in the [code ](https://github.com/lightninglabs/aperture/blob/master/l402/header.go#L32) from Aperture. Sulu does the [same](https://github.com/sulusolutions/l402-ts/blob/e08be22ceec5a5c2ab5733de9c2dcb0efa212707/src/client/axios.ts#L41).

I think the use of `"` in the www-authenticate header, while to my knowledge not strictly necessary, its actually quite typical. See for example:
- [RFC 6750](https://datatracker.ietf.org/doc/html/rfc6750#section-3) Bearer Auth Scheme
- [RFC 7617](https://datatracker.ietf.org/doc/html/rfc7617#section-2) Basic Auth Scheme

See also [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/WWW-Authenticate).

In my opinion we should keep with standard practices here, so I see no problem maintaining the `"` for the www-authenticate header.

positiveblue: You are right, I was following this diagram and checked the wrong header:

```
sequenceDiagram
     participant L as Lightning Node
     participant C as Client
     participant S as Server
     C->>S: GET HTTP/1.1 Request
     S->>C: HTTP/1.1 402 Payment Required
     Note over S: WWW-Authenticate: L402 macaroon=X invoice=Y
     C->>L: Pay invoice
     L->>C: Returns pre-image
     C->>S: GET HTTP/1.1 Request
     Note over C: Authorization: L402 macaroon:preimage
     S->>S: Verify L402 authorization
     S->>C: HTTP/1.1 200 OK
 ```


I see that [even aperture does it that way too](https://github.com/lightninglabs/aperture/blob/3a8936c4480425c0dd40a4b97864510a0fd78cc4/auth/authenticator.go#L110C2-L111C63) 

Case closed

**Comment by TheBlueMatt** [comment]
Any desire to pick this back up @Roasbeef? I recently saw someone avoid L402 and build their own thing because they thought they had to use macaroons, which they figured was overcomplicated so they just avoided it. It would be nice to have a real spec that clarifies that that is, in fact, not required.

File: blip-0026.md [hunk 41]
```diff
@@ -0,0 +1,641 @@
...
+   use"**_](https://tools.ietf.org/html/rfc7231#section-6.5.2)  **-- but this
+   document assumes the future has arrived.**
+
+4.  Alongside the 402 status code, the server MUST should specify the
+    `WWW-Authenticate` header \([_\[RFC7235\], Section
+    4.1_](https://tools.ietf.org/html/rfc7235#section-4.1)\) field to indicate
+    the L402 authentication scheme and the macaroon needed for the client to
+    form a complete L402.
+
+    A valid response header MUST take the form of:
```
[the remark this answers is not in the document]
Roasbeef: SGTM.

blip-0026.md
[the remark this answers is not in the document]
Roasbeef: Will update to mention it's BOLT 11. Whenever the bolt 12 invoice format becomes its own thing, we can update the language here to reflect the existence of both.

blip-0026.md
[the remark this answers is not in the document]
Roasbeef: The server sends it as base64 already. If the client knows what kind of macaroon/token was sent, then it can decode it itself to understand what capabilities it has.

Roasbeef:
> L402 and build their own thing because they thought they had to use macaroons, which they figured was overcomplicated so they just avoided it. It would be nice to have a real spec that clarifies that that is, in fact, not required.

Will make the requirements around what kind of API key is sent back more generic. Just curious: what did they reach for instead of macaroons? It's just an hmac chain, far simpler than say, something like [JWT](https://datatracker.ietf.org/doc/html/rfc7519) once you look under the covers.

TheBlueMatt:
I didn't ask. May well have just been an identifier to look up the invoice in a payment store later.

Roasbeef:
Aight just pushed a significant re-write of key sections. 

Biggest changes:
  * add version field as suggested above
  * sections condensed 
  * macaroon relaxed to just `token`, only requirement is that the token somehow commit to the preimage to allow for stateless server verification 

Updating our libraries to add backwards compat for the prior versions vs this.

blip-0026.md
```diff
@@ -0,0 +1,650 @@
...
+supporting server signals that a payment is required for an endpoint by sending
+a special `WWW-Authenticate` header along with a 402 status code. This header
+encodes an authentication token \(the L402 credential\), and a Lightning
+Network invoice. In order to traverse the L402-enabled endpoint, the client
+then pays the Lightning Network invoice, and presents the `(token, preimage)`
+tuple to the server in their next request via an HTTP `Authorization` header.
+Critically, the L402 token MUST cryptographically commit to the payment hash of
+the LN invoice sent by the server proxy. This commitment structure enables the
+server to easily verify that a client has properly paid for the endpoint in a
+stateless manner, providing in-band payment authentication.
```
positiveblue, line 57: > Critically, the L402 token MUST cryptographically commit to the payment hash of
the LN invoice sent by the server proxy. This commitment structure enables the
server to easily verify that a client has properly paid for the endpoint in a
stateless manner, providing in-band payment authentication.

I think we are constraining the protocol way too much here. From "proof of payment in the request" to "the proof of payment is the preiamge of a lightning invoice" to "the token contains all the information to be able to verify this in a stateless way"

blip-0026.md
```diff
@@ -0,0 +1,650 @@
...
+
+The L402 protocol is token-format agnostic: any authentication token that can
```
positiveblue, line 59: The same thing that "forced" use to say "token agnostic" should apply to "payment method agnostic"

blip-0026.md
```diff
@@ -0,0 +1,650 @@
...
+commit to a payment hash may be used. Macaroons (HMAC chain based bearer
+credentials) are the RECOMMENDED token format due to their support for
+delegation, attenuation of capabilities, and stateless verification. See the
```
positiveblue, line 62: > Macaroons (HMAC chain based bearer
credentials) are the RECOMMENDED token format due to their support for
delegation, attenuation of capabilities, and stateless verification.

This is great but orthogonal to the things a payment protocol/flow should care about

blip-0026.md
```diff
@@ -0,0 +1,650 @@
...
+
+The L402 protocol can be implemented as a simple reverse-proxy in front of a
+traditional HTTP/gRPC based backend. The proxy handles the minting +
+verification of the L402 credentials, decoupling the authentication layer from
+the target web service backends. In more advanced deployments, a backend may
+introspect into the L402 token structure to enforce capability verification,
+and even dynamically communicate with the proxy to dictate real-time pricing in
+an automated fashion.
+
+The L402 protocol gets its name from the HTTP status code 402: Payment
```
positiveblue, line 74: <3

blip-0026.md
```diff
@@ -0,0 +1,650 @@
...
+stateless manner, providing in-band payment authentication.
+
+The L402 protocol is token-format agnostic: any authentication token that can
+commit to a payment hash may be used. Macaroons (HMAC chain based bearer
+credentials) are the RECOMMENDED token format due to their support for
+delegation, attenuation of capabilities, and stateless verification. See the
+[Appendix](#appendix) for a full specification of the macaroon-based token
+format.
+
+The L402 protocol can be implemented as a simple reverse-proxy in front of a
```
positiveblue, line 66: This assumes "all request need to go through here"

The current flows do not follow this at all.

A) I do not handle payments and delegate to a third party SOMETIMES (only when a payment is involved) and that third party tells me when a payment was processed (e.g: stripe)
B) I check if a payment was received (polling my Lightning node, checking a chain, pulling the latest transactions in my bank account, etc...)

Also, most of the time I have an state in the service (subscribed as Free, Premium, Pro or a number of credits etc...)

The payment changes the state but the client SHOULD NOT have to show a proof of payment every time

positiveblue (commented):
I gave it another read. All feedback is based on building real systems that have the problems L402 is trying to solve + feedback from developers who said no to this solution.

However, it may be more about differences about what to build than feedback for this specific proposal so feel free to ignore.


L402 seems to have a really specific (bounded) vision of the future where:

- Payments MUST happen in the LN
- Ticket-like credentials the server can (and MUST) verify every time
- Tickets are enabled only after payment (so they can provide a proof of payment)
- API Gateway/middleware for all requests in the service


I think what would be more useful (and help adoption) would be something with way less constraints that still allows us to create systems like the ones proposed here but does not require people change everything to build services that way.

blip-0026.md
TheBlueMatt: Yea, this isn't required to implement it. The server could validate the payment on its end if it wants (I mean it shouldn't, but it definitely can). There's no need to specify this, let alone in an "abstract".

blip-0026.md
TheBlueMatt: Rewriting this to be "payment method agnostic" seems way out of scope. The point of this document is to describe a thing that is already deployed, not to invent a new payment protocol.

blip-0026.md
```diff
@@ -0,0 +1,641 @@
...
+offer a new tier of paid APIs that sit between free, and subscription: pay as
+you go.
+
+One can view L402 credential (a macaroon) as a fancy authentication credential
+or cookie. They differ from regular cookies in that they're a cryptographically
+verifiable bearer credential. An L402 credential _encodes_ all its capabilities
+within a macaroon which can only be created by the end service provider. The
+L402 specification uses a combination of `HTTP` as well as the Lightning Network
+to create a seamless end-to-end payment+authentication flow for the
+next-generation of paid APIs built on top of the Lightning Network.
```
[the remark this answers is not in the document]
TheBlueMatt: This wasn't addressed.

blip-0026.md
```diff
@@ -0,0 +1,650 @@
...
+
+One can view an L402 credential as a fancy authentication credential or cookie.
+They differ from regular cookies in that they're a cryptographically verifiable
+bearer credential. An L402 credential _encodes_ all its capabilities within a
+token which can only be created by the end service provider. The L402
+specification uses a combination of `HTTP` as well as the Lightning Network to
+create a seamless end-to-end payment+authentication flow for the
+next-generation of paid APIs built on top of the Lightning Network.
+
+## Example Applications and Use Cases
```
TheBlueMatt, line 105: I'm not sure what the value of this section is after the abstract.

blip-0026.md
```diff
@@ -0,0 +1,650 @@
...
+    S->>C: HTTP/1.1 200 OK
+```
+
+HTTP-based protocol flow:
+
+1. A client hits an `HTTP` endpoint to fetch a resource.
+
+2. The server decides that payment is required to access the endpoint. They
+   return a HTTP status code of `402` along with a `WWW-Authenticate` header
+   containing a token and invoice. The token MUST commit to the payment hash
```
TheBlueMatt, line 169: again this doesn't need to be specified.

blip-0026.md
```diff
@@ -0,0 +1,650 @@
...
+   of the invoice `H`.
+
+3. The client pays the invoice over the Lightning Network and receives the
+   payment pre-image `r` to the payment hash `H`, `H = sha256(r)`
+
+4. The client re-issues the request with an `Authorization` header that
+   includes the payment preimage `r`, and the token sent by the server.
+
+5. The server cryptographically verifies the authenticity and integrity of the
```
TheBlueMatt, line 178:
```suggestion
5. The server verifies the authenticity and integrity of the
```

blip-0026.md
```diff
@@ -0,0 +1,650 @@
...
+   token, then ensures the payment hash committed to in the token `H`,
+   satisfies the relation `H = sha256(r)`.
+
+6. The server returns a 200 OK status code along with the specified resource.
+
+If the service supports persistent L402 token re-use, then a client SHOULD
```
TheBlueMatt, line 184: How does a client know to do this? Its not signaled. Presumably the client should always do it and if they get another 402 they have to pay again.

blip-0026.md
```diff
@@ -0,0 +1,650 @@
...
+tier upgrade, and potentially others \(discussed further in the higher-level
+design document\).
+
+L402 could be configured for use on a per-backend-service basis or for all
+exposed services. As an example, the same L402 credential may commit to
+capabilities enabling a user to access both service `Y` and `X` from the same
+endpoint. This flexibility is afforded because all services are going to be
+gated by the same L402 proxy, which verifies all credentials for all backend
+services.
+
```
TheBlueMatt, line 205: Not sure what any of this adds. Yes, the service can issue different tokens for different things or a single token for different things.

blip-0026.md
```diff
@@ -0,0 +1,650 @@
...
+transmission is not encrypted in some way such as TLS, the L402 can be used by
+Mallory and the L402 proxy would not be able to distinguish this usage as
+illicit.
+
+L402 authentication is also vulnerable to spoofing by counterfeit servers. If a
+client slightly mistypes the URL of a desired backend service, they become
+vulnerable to spoofing attacks if connecting to a server that maliciously
+stores their L402 and uses it for their own purposes.
+
+Because L402 tokens support attenuation through caveats, this class of attack
```
TheBlueMatt, line 218: No they dont, they're generic now.

blip-0026.md
```diff
@@ -0,0 +1,650 @@
...
+context. Each binding predicate carries its own tradeoffs: an IP caveat
+prevents use after a network change, while a TLS client cert fingerprint
+requires the client to maintain a stable key pair. Deployments should choose
+binding predicates appropriate to their threat model.
+
+## L402 HTTP/gRPC Protocol Specification
+
+### HTTP Protocol Flow Specification
+
+In this section, we specify the protocol for the HTTP portion of the L402
```
TheBlueMatt, line 234: Why is this even separate from the client spec? Its the same thing...

blip-0026.md
```diff
@@ -0,0 +1,650 @@
...
+
+4. Alongside the 402 status code, the server MUST specify the
+   `WWW-Authenticate` header \([_\[RFC7235\], Section
+   4.1_](https://tools.ietf.org/html/rfc7235#section-4.1)\) field to indicate
+   the L402 authentication scheme and the token needed for the client to form a
+   complete L402.
+
+   A valid response header MUST take the form of:
+   ```
+   WWW-Authenticate: L402 version="0", token="T", invoice="P"
```
TheBlueMatt, line 272: Why doesn't this appear in the client section?

blip-0026.md
```diff
@@ -0,0 +1,650 @@
...
+     required to complete the `Authorization` header.
+
+   An example L402 HTTP response resembles:
+
+   ```text
+   HTTP/1.1 402 Payment Required
+
+   Date: Mon, 04 Feb 2014 16:50:53 GMT
+
+   WWW-Authenticate: L402 version="0", token="AGIAJEemVQUTEyNCR0exk7ek90Cg==", invoice="lnbc1500n1pw5kjhmpp5fu6xhthlt2vucmzkx6c7wtlh2r625r30cyjsfqhu8rsx4xpz5lwqdpa2fjkzep6yptksct5yp5hxgrrv96hx6twvusycn3qv9jx7ur5d9hkugr5dusx6cqzpgxqr23s79ruapxc4j5uskt4htly2salw4drq979d7rcela9wz02elhypmdzmzlnxuknpgfyfm86pntt8vvkvffma5qc9n50h4mvqhngadqy3ngqjcym5a"
```
TheBlueMatt, line 296: Wait, did we rename `macaroon` to `token` in the wire encoding? That would break existing impls right?

blip-0026.md
```diff
@@ -0,0 +1,650 @@
...
+   itself, or pass it on to the proxied backend.
+
+Regarding step 3.1 and 3.2, it's imperative that the server ensure that the
+invoice has been paid in full before processing the response, or proxying it to
+a backend. As stated above, by cryptographically committing to the payment
+hash, the server can implement fast stateless verification of the payment
+hash+preimage relation.
+
+
+#### Client Flow
```
TheBlueMatt, line 333: We already specified the client above!

blip-0026.md
```diff
@@ -0,0 +1,650 @@
...
+Upon receiving a valid L402 `WWW-Authenticate` header the client:
+
+1. Should verify that the contained BOLT 11 invoice doesn't request an
+   excessive amount of Bitcoin.
+   1. If so, then the client SHOULD abandon the HTTP request.
+2. After validating the returned BOLT 11 invoice, the client MUST pay the
+   invoice over the LN to receive the payment pre-image `r`.
+3. The client then constructs a valid `Authorization` header with the
+   following form:
+   `Authorization: L402 <base64(token)>:<hex(preimage)>`
```
TheBlueMatt, line 344: I assume you don't intend the client to base64-encode the token it already received?

TheBlueMatt (commented):
This still needs a lot of work. There's a ton of extraneous context and repetitive text or text that describes specific scenarios that should be in an appendix, if included at all. The specification section also seems inconsistent in a few places and is also very redundant.

Roasbeef:
@positiveblue 

> L402 seems to have a really specific (bounded) vision of the future where:
> Payments MUST happen in the LN


Indeed that's the case. There's likely room to have a variant where even something like credit cards are accepted. However that doesn't fit the vision we're pursuing here. I believe that the Lightning Network has a number of unique qualities compared to other payment networks that make it well suited for the application of low-latency+scalable machine to machine payments. 

I currently have no desire to expand the spec to support other payment networks, as that would just bloat the spec, and make it less focused. Ofc you're free to experiment in the space, I look forward to learning from your experiences in that endeavor.

blip-0026.md
Roasbeef: Example use cases.

blip-0026.md
Roasbeef: That the token must commit to the payment hash? It's one of the key aspects of the protocol that enables ~stateless verification!

blip-0026.md
Roasbeef: Few ways:
  * They're able to introspect into the token itself, and can ascertain w/e their remaining allotment is. 
  * They try again, and get another 402, GOTO 1. 
  * Server can signal some other manifest page, or have an `/info` endpoint (or w/e) where it returns structure info re the submitted token.

blip-0026.md
Roasbeef: >  As an example, the same L402 credential may commit tocapabilities enabling a user to access both service `Y` and `X` from the same endpoint.

Fine grained auth.

blip-0026.md
Roasbeef: See backwards compat section.

Roasbeef:
So we've recently updated our [L402 github page](https://github.com/lightninglabs/l402). We plan on continuing development of the protocol there. IMO a stand alone repo is better suited as it's a targeted area that devs can submit issue+PRs and/or utilize the Discussions feature to collaborate or ideate. 

We can also provide additional supporting documentation such: as the [mini agent spec](https://github.com/lightninglabs/L402/blob/master/agent-spec.md), a [more structured description of macaroons](https://github.com/lightninglabs/L402/blob/master/macaroon-spec.md) for those that opt to use them, future agent oriented Skills, etc. Reviving the spec repo also allow us to adapt and move more quickly, eg: adding future extensions in a new version to support `offer=`, etc. 

This PR can stay open, or be merged, but we'll focus our efforts on the repo above as it was the original written spec/site/docs. 

We invite all interested devs to contribute+collaborate at the [L402 Github](https://github.com/lightninglabs/l402)!

PR #28, opened 2023-09-05
Introduce Paratonnerre

t-bast:
Paratonnerre is a protocol to safely control a remote lightning node from an untrusted application using a secure hardware device.


rustyrussell:
Hmm, like a reduced-subset commando! (Yay, no JSON!).

Types are 16 bit, so your numbers don't work?

I prefer bearer tokens (runes/macaroons) but that does require more setup, so simple nodeid auth makes sense.

Will review in more depth later...

t-bast:
The actual commands definitely need more work, I probably need to spend a bit more time on them (and implement that in `eclair` to make sure nothing odd comes out).

The goal is really to only use this tool for commands that involve spending some bitcoin, because that's what you really want to secure as much as possible. All the "read" commands (listing channels, getting information about your node status) don't make sense here in my opinion (no need for HSM security for that), but we may still want to add them as optional commands at some point.

I'm not sure yet what would be best for "feature negotiation" (agreeing on which commands a specific `node_id` is allowed to send), it looks like this could simply be configured on the lightning node side and exchanged during `init`, but may be worth thinking about.

PR #29, opened 2023-09-06
blip-tap: initial bLIP draft for Taproot Asset Protocol channels

Roasbeef:
This bLIP describes a variant on the "simple taproot channels" proposal that also supports holding an transferring assets created by the Taproot Assets Protocol. As Taproot Assets are built on top of the taproot itself, from the PoV of the taproot channel format, Taproot Assets manifests entirely as an extra tapscript sibling placed in the tapscript tee of relevant outputs. A set of asset-specific balances (in the form of taproot asset tree commitments) are maintained as an overlay layer on top of the normal initiator+responder balances of Lightning channels. For channel state transitions and eventual on-chain contract claims, in addition to normal taproot witnesses, a set of taproot asset level witnesses are also exchanged, encumbered by a nested iteration of the current Tapscript VM, the Taproot Assets VM.

In order to facilitate multi-hop payments of the existing LN using Taproot Assets edge liquidity, an RFQ (Request For Quote) last-mile negotiation scheme is used to lock in an exchange rate for both incoming and outgoing payments by liquidity providers. Tendered quotes `(asset_id, volume, price)` are identified by a cryptographic hash and scid-like sequence number, and ephemerally expire in order to reduce exchange rate risk. The existing BOLT 11 invoice format is used verbatim, in a manner that allows a receiver to accept an taproot asset without burdening the sender with up to date knowledge of exchange rates. As no effective changes to the invoice scheme re required, support for BOLT 12 invoices is readily available.


blip-tap.md
```diff
@@ -0,0 +1,1194 @@
...
+affectively unbounded amount of assets to be transferred in a single transaction.
+
+#### Asset Splits
+
+Similar to normal Bitcoin UTXOs, UTXOs for Taproot Assets can also be split and
+merged. When splitting an asset, in order to ensure that the supply of an asset
+is conserved (no inflation occurred) any resulting UTXOs created from the
+source UTXO are inserted into a special MS-SMT tree dubbed a
+`split_commitment`. A "root" asset is designated which holds the root hash of
+the `split_commitment`. Any splits that were created from the root asset
```
guggero, line 215: ...created from the root asset are committed into the split commitment tree? Or how should the sentence end?

blip-tap.md
```diff
@@ -0,0 +1,1194 @@
...
+
+Throughout this document, we assume that `option_taproot` was negotiated, and
+also the `option_taproot` channel type is used.
+
+### New TLV Types
+
+Note that these TLV types exist across different messages, but their type IDs
+are always the same.
+
+The starting TLV type for these non-official values is `75249`.
```
guggero, line 339: Any rationale behind this number? Or just random?

blip-tap.md
```diff
@@ -0,0 +1,1194 @@
...
+the responder that the asset exists, and that the initiator is capable of
+moving them. The `asset_proof` itself is fully singed, which then allows the
+responder to independently arrive at the final TAP asset root to be committed
+into the funding output.
+
+**Requirements**
+
+The sending node:
+
+* MUST send a new `tap_asset_proof` for each asset UTXO they wish to anchor in
```
guggero, line 413: So this is simply for the case where the funding transaction would merge multiple asset inputs of the same asset ID into one? But the channel funding output would only commit to a single (merged) asset UTXO, right?

blip-tap.md
```diff
@@ -0,0 +1,1194 @@
...
+
+The `tx_asset_proof` message links a prospective channel funding attempt to a
+set of proposed assets via the `temporary_channel_id`. For each asset UTXO that
+is to be anchored in the final `musig2` funding output, a new `tx_asset_proof`
+message is sent.
+
+Each `tap_asset_proof` declares an `asset_id`, an `amt`, and finally a
+serialized existence proof for the Taproot Asset.
+
+1. type: ?? (`tap_asset_proof`)
```
guggero, line 393: So if there can be multiple asset inputs, would the sending node send multiple `tap_asset_proof` objects? Or should we add a `num_proofs` to this type to allow multiple proof files to be sent? Because the requirement of all sharing the same `asset_id` would be enforced by that.

guggero: Ah, I only now saw that `tx_asset_proof` is its own _message_ not part of the `open_channel` message... What's the reason for that? Couldn't the `open_channel` message just contain a list of proofs? Or are there overall message size limits?

blip-tap.md
```diff
@@ -0,0 +1,1194 @@
...
+  message
+* MUST set the `tap_partial_signature_with_nonce` TLV type to be the `musig`
+  partial signature of the TAP funding output as specified in the section on
+  Funding Output Creation
+  * This signature will cover the TAP virtual tx that sending the TAP funding
+    output for the TAP commitment transaction
+
+The receiving node:
+
+* MUST verify the incoming partial signature against the constructed TAP
```
guggero, line 498: Do we need to mention that the receiving node MUST check that all anchor transactions referenced by the proofs are actually present as inputs to the funding transaction?

blip-tap.md
```diff
@@ -0,0 +1,1194 @@
...
+  force close transaction with the commitment transaction
+
+#### Funding Output Construction
+
+The funding output for the TAP layer is effectively a direct mirror of the
+funding output on the Bitcoin layer, with a normal tapscript tweak used rather
+than a BIP 86 tweak.
+
+The `tap_asset_proof` message sends all the fully singed asset TLV leafs ahead
+of time, as
```
guggero, line 542: nit: incomplete sentence.

blip-tap.md
```diff
@@ -0,0 +1,1194 @@
...
+
+// TODO(roasbeef): partial reveal here? needed in any case to accept funding,
+but has similar issues to dual funding reveal, but constrained
+
+Given a set of fully signed TAP asset leaf TLVs (each dubbed `tap_input_leaf`),
+the funding output is constructed as follows:
+
+  1. Initialize an empty TAP asset tree dubbed `tap_asset_tree`
+
+  2. For each `tap_input_leaf`:
```
guggero, line 552: What do you mean by **for each** `tap_input_leaf` here? Wouldn't there only be a single one to reduce complexity? Meaning that even if there are multiple inputs, they would be merged into a single asset level UTXO?

blip-tap.md
```diff
@@ -0,0 +1,1194 @@
...
+    valid split for the balance of the responder.
+
+The recipient:
+  - MUST verify the `tap+partial_signature` field using the
+    `PartialSigVerifyInternal` algorithm of `bip-musig2` over the specified
+    `tap_virtual_tx` :
+  
+    - if the partial signature is invalid, MUST fail the channel
+
+### Channel Operation
```
guggero, line 616: Checkpointing my review here, will continue tomorrow.

guggero (commented):
Awesome read so far!
Checkpointing my review, will continue later.

I'm fixing typos directly in a branch (https://github.com/guggero/blips/tree/taproot-assets-blip-typos), feel free to pull in directly (to reduce the amount of nit related chatter on the PR).

blip-tap.md
```diff
@@ -0,0 +1,1194 @@
...
+to_local_output_key = taproot_nums_point + tagged_hash("TapTweak", taproot_nums_point || to_delay_script_root)
+```
+
+Note how the existing scripts (`to_delay_script`, and `revoke_script`) are used
+as a nested sub-list to create the `to_delay_script_root`. This is due to the
+fact that for TAP asset, the tapscript commitment needs to be the only element,
+or the highest element in the tree. The spec of `bip-tap.mediawiki` elaborates
+on this.
+
+The `tap_to_local_script_root` is itself, a nested instance of the _existing_
```
guggero, line 677: This sounds a bit confusing or unclear. You're meaning to say that `tap_to_local_script_root` is the root of two levels of MS-SMT trees with a leaf at the very bottom that has the same `to_delay_script_root` construction as its `script_key`, right?

blip-tap.md
```diff
@@ -0,0 +1,1194 @@
...
+#### Accepting Quotes  (`tap_rfq_accept`) 
+
+If it can, then it should send `tap_rfq_accept` that returns the quote amount
+the edge node is willing to observe to move `N` units of asset `asset_id`:
+
+1. type: ?? (`tap_req_accept`)
+2. data:
+     * [`32*byte`:`rfq_id`]
+     * [`BigSize`:`accepted_rate_tick]
+     * [`BigSize`:`expiry_seconds`]
```
guggero, line 991: Do we need `BigSize` for seconds? Or does that just mean `VarInt`?

blip-tap.md
```diff
@@ -0,0 +1,1194 @@
...
+
+When a receiver wishes to receive `N` units of TAP asset ID `asset_id`, a new
+p2p message `tap_rfq` is sent with the following structure:
+
+1. type: ?? (`tap_rfq`)
+2. data:
+     * [`32*byte`:`rfq_id`]
+     * [`32*byte`:`asset_id`]
+     * [`BigSize`:`asset_amt`]
+     * [`BigSize`:`suggested_rate_tick`]
```
guggero, line 932: Do we need to add the short channel ID of the channel to the quote request/response?
It could be that we have two USD backed channels, would the quote be valid for both?
Because on the last hop we only have this fake short channel ID that binds to the quote. But if there are multiple channels that could satisfy the quote, will the last hop just choose one of them?

blip-tap.md
```diff
@@ -0,0 +1,1194 @@
...
+* `asset_id` is the asset ID of the asset the receiver intends to receive
+* `asset_amt` is the amount of units of said asset
+* `suggested_rate_tick` is the internal unit used for asset conversions. A tick
+  is 1/10000th of a currency unit. It gives us up to 4 decimal places of
+  precision (0.0001 or 0.01% or 1 bps). As an example, if the BTC/USD rate was
+  $61,234.95, then we multiply that by 10,000 to arrive at the `usd_rate_tick`:
+  `$61,234.95 * 10000 = 612,349,500`.  To convert back to our normal rate, we
+  decide by `10,000` to arrive back at `$61,234.95`.
+
+Given valid `rfq_id`, we then define an `tap_rfq_scid` by taking the last `8`
```
guggero, line 947: Do we need to make sure that a derived `tap_rfq_scid` never collides with an actual SCID we use for any of our local channels? Otherwise we couldn't tell the two meanings apart from each other.

guggero (commented):
Great stuff! I really love the elegance of the quote system and the SCID alias use. Really makes it easy to do exchange rates at either or both ends of a payment.

Will need a second pass to catch all the low-level details, but at least think I have a decent understanding of the flow and data exchanged to make this work.

blip-tap.md
```diff
@@ -0,0 +1,1194 @@
...
+### Last Mile Routing
+
+In order to bridge channels with TAP assets to the Lightning Network, a
+specialized set of last-mile liquidity providers are required. These providers
+have incoming/outgoing channels of TAP assets, then utilize the _existing_ HTLC
+mechanism to complete the route by bridging the TAP edge liquidity with the
+Bitcoin Backbone portion of the Lightning Network.
+
+For each logical payment to be accepted, the Edge Node and the Liquidity
+Provider engage in Request-For-Quote (RFC) swap negotiation. Once a price has
```
alexlwn123, line 904: Typo? RFC vs. RFQ

`tap_rfc_hash`, `rfc_id`, etc.

blip-tap.md
```diff
@@ -0,0 +1,1194 @@
...
+
+  - MUST set `rfq_id` to the matching `rfq_id` sent in a prior `tap_rfq` message
+
+  - MUST set `accepted_rate_tick` to a value they deem to be an acceptable
+    exchange rate
+
+  - MUST set `expiry_seconds` to the relative expiry time in the future that
+    the quote will expire after which
+
+  - MUST set `rfg_sig` to be a BIP-340 schnorr signature over the serialized
```
alexlwn123, line 1020: typo `rfg_sig` -> `rfq_sig`

blip-tap.md
Roasbeef: Fixed, basically that proves no assets were inflated.

blip-tap.md
Roasbeef: I think I had GPT do something like sum everything mod some number? kek

blip-tap.md
Roasbeef: > So this is simply for the case where the funding transaction would merge multiple asset inputs of the same asset ID into one?

Yeah, this is the phase where the initiator proves that the assets actually exist, and if they're merging N assets into a single UTXO. Correct that it'll still be a single merged UTXO. 

This is also has a play for the eventual addition of dual funding as well.

blip-tap.md
Roasbeef: > So if there can be multiple asset inputs, would the sending node send multiple tap_asset_proof objects? 

Yeah if we have 3 inputs of the same asset, we send 3 proofs. If we have 2 diff assets, we send two proofs, etc. 

> Or should we add a num_proofs to this type to allow multiple proof files to be sent?

I made them each a single message so we wouldn't have to worry about overflowing the default 65 KB message size. At least in the early days. The idea here is this is _just_ the anchor proof of the final resting place, and not the entire history proof. Using this, then the verifier would use a universe to fetch the full history or up to w/e check point. 

> What's the reason for that? Couldn't the open_channel message just contain a list of proofs? Or are there overall message size limits?

I had them a distinct messages due to size limits, and also to minimize the changes needed for `open_channel`.

blip-tap.md
Roasbeef: So in single funder land, the initiator never actually see the full funding transaction. This would be something done once the funding transaction has been confirmed, or if we modified things to just have the initiator send directly to the responder.

blip-tap.md
Roasbeef: Heh, I think this is the moment I realized we'd want a segwit like encoding. Pushed a fixup to fill out the commit.

blip-tap.md
Roasbeef: Correct that in the default case (only one UTXO in the TAP tree), we'd only need one. This is for the case of multiple  assets in the funding output. If it's just a single asset, even if we send in multiple inputs, then correct that we'll only have a single item in the TAP tree of the funding output.

blip-tap.md
Roasbeef: What I mean here is that we basically replicate all the scripts on the TAP layer. Tacked on an extra sentence to help clarify a bit.

blip-tap.md
Roasbeef: Yeah it's basically just a `VarInt`, but big-endian, as the Bitcoin varint is little-endian.

blip-tap.md
Roasbeef: > Do we need to add the short channel ID of the channel to the quote request/response?


Hmm, good question. I think for the MPP case, we'd want to treat the set of USD backed channels as a single unit? Since here we're concerned with accepting a certain price for a a given volume (amt) transferred. 

Today we have "non-strict forwarding" by default, so the last hop can always choose _which_ channels to actually send over, and this is the default implementation in the switch. 

I think this is something we'll want to come back to though.

blip-tap.md
Roasbeef: Yeah the'll need to be distinct. Today the alias manager uses a block height in the _past_ before segwit was even activated, so it won't collide until we make hundreds of thousands of aliases. However as invoices expire, we can garbage collect old aliases for reuse. 

I added this as a requirement.

blip-tap.md
Roasbeef: Nice catch! Fixed.

blip-tap.md
Roasbeef: Ah yeh, habit, should be RFQ.

Roasbeef:
Thanks for the review so far! Pushed up a fixup commit implementing changes/clarifications.

blip-tap.md
```diff
@@ -0,0 +1,1200 @@
...
+specialized set of last-mile liquidity providers are required. These providers
+have incoming/outgoing channels of TAP assets, then utilize the _existing_ HTLC
+mechanism to complete the route by bridging the TAP edge liquidity with the
+Bitcoin Backbone portion of the Lightning Network.
+
+For each logical payment to be accepted, the Edge Node and the Liquidity
+Provider engage in Request-For-Quote (RFQ) swap negotiation. Once a price has
+been accepted, a final response is returned that uniquely identifies that
+specified price. The final quote is then identified by either a `tap_rfq_scid`
+or a `tap_rfq_hash`. The former is laced in the first hop for an outgoing
```
GeorgeTsagk, line 910: `tap_rfq_hash` is only mentioned once here, was there an intention for it?

blip-tap.md
```diff
@@ -0,0 +1,1200 @@
...
+payment with a TAP origin, and the latter placed within the normal onion
+payload for a last-mile hop into a TAP asset.
+
+All quotes are ephemeral and will expire along side the created invoice.
+Epidermal quotes ensure that all sides are able to limit their exposure to
+price fluctuations. It's recommended that invoices expire within minutes in
+order to clamp down on price volatility exposure.
```
GeorgeTsagk, line 917: I understand that invoices can expire that fast, but this kind of short-timeout can't be expressed in the HTLCs?

blip-tap.md
```diff
@@ -0,0 +1,1200 @@
...
+* `rfq_id` is a randomly generate 32-byte value to uniquely identify this RFQ
+  request
+* `asset_id` is the asset ID of the asset the receiver intends to receive
+* `asset_amt` is the amount of units of said asset
+* `suggested_rate_tick` is the internal unit used for asset conversions. A tick
+  is 1/10000th of a currency unit. It gives us up to 4 decimal places of
+  precision (0.0001 or 0.01% or 1 bps). As an example, if the BTC/USD rate was
+  $61,234.95, then we multiply that by 10,000 to arrive at the `usd_rate_tick`:
+  `$61,234.95 * 10000 = 612,349,500`.  To convert back to our normal rate, we
+  decide by `10,000` to arrive back at `$61,234.95`.
```
GeorgeTsagk, line 948:
```suggestion
  divide by `10,000` to arrive back at `$61,234.95`.
```

blip-tap.md
```diff
@@ -0,0 +1,1200 @@
...
+
+The recipient:
+
+  - MUST not attempt to send/receive using the rejected `rfq_id`
+
+
+#### First Hop TAP HTLC Onion Processing
+
+When sending a payment that uses an outbound TAP asset HTLC as it's origin, in
+order for the payment succeed, the sender must place the negotiated `rfq_id` in
```
GeorgeTsagk, line 1069:
```suggestion
order for the payment to succeed, the sender must place the negotiated `rfq_id` in
```

blip-tap.md
```diff
@@ -0,0 +1,1200 @@
...
+
+In the event that the last hop in a route receive a payload that has an
+`short_channel_id` value that matches a prior accepted `rfq_scid` value, then
+this indicates that the sender is attempting to pay a receiver in the asset
+bound by the `rfq_id` and `rfq_scid`.
+
+Note that we don't require that the sender use any special values other than
+what is already known in the existing protocol. This enables _unupgraded_
+senders to send BTC, with the receiver obtaining their asset of choice, without
+the sender needing to worry about exchange rates at all.
```
GeorgeTsagk, line 1128: :+1:

blip-tap.md
```diff
@@ -0,0 +1,1200 @@
...
+
+When receieving an incoming onion payload with a known `rfq_scid` value, the
+receiver:
+
+- MUST reject the HTLC is `tap_scid` is expired based on the posted
```
GeorgeTsagk, line 1133:
```suggestion
- MUST reject the HTLC if `tap_scid` is expired based on the posted
```

blip-tap.md
```diff
@@ -0,0 +1,1200 @@
...
+  `expiry_seconds` value
+
+- MUST reject the entire HTLC set if at anytime, the sum of HTLCs (the
+  `amt_to_forward` field) targetting `tap_rfq_scid` eceeds the negotiated
```
GeorgeTsagk, line 1137:
```suggestion
  `amt_to_forward` field) targetting `tap_rfq_scid` exceeds the negotiated
```

blip-tap.md
```diff
@@ -0,0 +1,1200 @@
...
+      tick rate of `612_349_500` ($61,234.95), then the penultimate node should
+      send $1000 to the last hop:
+        * `amt_to_forward = ((incoming_amt_msat * accepted_rate_tick) // msat_multiplier) / tick`
+        * `amt_to_forward = ((1_633_054_326 * 612_349_500) // 100_000_000_000) / 10_000`
+        * `amt_to_forward = ((999999999998937000) // 100_000_000_000) / 10_000`
+        * `amt_to_forward = (9999999 / 10_000)`
+        * `amt_to_forward = (9999999 / 10_000)`
+        * `amt_to_forward = 999.9`
+    * Note that all assets internally are accounted in a unit of a `tick`
+      (1/1000th) of an asset. When convering back to the main asset, the value
```
GeorgeTsagk, line 1156:
```suggestion
      (1/1000th) of an asset. When converting back to the main asset, the value
```

blip-tap.md
```diff
@@ -0,0 +1,1200 @@
...
+widely deployed across the Lightning Network. During the negotiation process,
+after a `tap_rfq_scid` is accepted, the border node then will forward received
+HTLCs to that final hop.
+
+From the PoV of the sender, the route construction when starting from a
+BTC-only link is no different. The final rate negotiated is then transparently
+passed on as an additional last-hop fee.
+
+When creating an invoice, the creator MUST ensure that the invoice expiry value
+is set exactly to the `expiry_seconds` value of the accepted RFQ.
```
GeorgeTsagk, line 1175: makes sense that both need to expire fast, but why on the exact same time?
e.g what if quote expired 10ms before invoice and vice versa

blip-tap.md
```diff
@@ -0,0 +1,1200 @@
...
+    is `612_349_500` or `$61,234.95`. To arrive at the `mSAT` value they should
+    put into the invoice:
+      * `invoice_amt = (asset_amt * tick * msat_multiplier) / accepted_rate_tick`
+      * `invoice_amt = (100 * 10_000 * 100_000_000_000) / 612_349_500`
+      * `invoice_amt = (100000000000000000) / 612_349_500`
+      * `invoice_amt = (100000000000000000) / 612_349_500`
+      * `invoice_amt = 163305432 mSAT`
+      * `invoice_amt = 163305 SAT`
+    
+Always expressing the invoice amount in BTC/mSAT ensures that unpugraded
```
GeorgeTsagk, line 1192:
```suggestion
Always expressing the invoice amount in BTC/mSAT ensures that unupgraded
```

blip-tap.md
```diff
@@ -0,0 +1,1200 @@
...
+senders will be able to send over these asset channels.
+
+
+## Universality
```
GeorgeTsagk, line 1196: what is this section going to be about?

GeorgeTsagk (commented):
Great write-up. Spotted a bunch of typos, also have a few Qs :carrot:

guggero (commented):
Some drive-by comments for stuff I noticed while implementing a mini-PoC.

blip-tap.md
Roasbeef: What use is the quote if the invoice has expired? Same with the other way around. 

Or is your idea that quotes can span multiple invoices? As is, they're considered to be 1:1

blip-tap.md
```diff
@@ -0,0 +1,1200 @@
...
+    `expiry_seconds` has elapsed
+
+#### Rejecting Quotes  (`tap_rfq_reject`) 
+
+In the event that an edge node is unable to satisfy a quote request, then they
+should send `tap_rfq_reject`, identifying the rejected quote ID. A quote might
+be rejected if the channel cannot accommodate the proposed volume, or if the
+edge node is unwilling to carry any HTLCs for that `asset_id`.
+
+1. type: ?? (`tap_req_accept`)
```
ffranr, line 1047: `tap_req_accept` -> `tap_rfq_reject`

blip-tap.md
```diff
@@ -0,0 +1,1200 @@
...
+
+* `rfq_id` matches the existing `rfq_id` of a set `tap_rfq`
+
+* `accepted_rate_tick` is the proposed rate for the volume unit expressed in
+  the internal unit of a `tick`.
+
+* `expiry_seconds` is the amount of seconds to use for the expiry of both the
+  quote and the invoice
+
+* `rfq_sig` is a signature over the serialized contents of the message
```
ffranr, line 1012: I'm not sure if this signature is necessary. Is the purpose of this field to facilitate distributing the message among peers?

blip-tap.md
```diff
@@ -0,0 +1,1200 @@
...
+
+#### Accepting Quotes  (`tap_rfq_accept`) 
+
+If it can, then it should send `tap_rfq_accept` that returns the quote amount
+the edge node is willing to observe to move `N` units of asset `asset_id`:
+
+1. type: ?? (`tap_req_accept`)
+2. data:
+     * [`32*byte`:`rfq_id`]
+     * [`BigSize`:`accepted_rate_tick]
```
ffranr, line 996: missing: `

blip-tap.md
```diff
@@ -0,0 +1,1200 @@
...
+
+Once the edge node has received, the `tap_rfq` message, then it should decide
+if it's able to accommodate the quote or not. 
+
+#### Accepting Quotes  (`tap_rfq_accept`) 
+
+If it can, then it should send `tap_rfq_accept` that returns the quote amount
+the edge node is willing to observe to move `N` units of asset `asset_id`:
+
+1. type: ?? (`tap_req_accept`)
```
ffranr, line 993: In choosing a type number we might want to consider doing the following:

Custom message types start at `32768`. We can specify a taproot-assets specific offset from that starting number by concatenating the alphabetical index positions of the letters "t" (20), "a" (1), and "p" (16), which gives `20116`. And then the tap offset is `TapMessageTypeBaseOffset = 32768 + 20116`.

Using that offset, the quote request message can have type `TapMessageTypeBaseOffset + 0`, quote accept with `TapMessageTypeBaseOffset + 1`, and quote reject with `TapMessageTypeBaseOffset + 2`.

What do you guys think?

blip-tap.md
```diff
@@ -0,0 +1,1200 @@
...
+2. data:
+     * [`32*byte`:`rfq_id`]
+     * [`BigSize`:`accepted_rate_tick]
+     * [`BigSize`:`expiry_seconds`]
+     * [`64*byte`:`rfq_sig`]
+
+TODO(roasbeef): tlv err where?
```
ffranr, line 1000: Perhaps we can include an optional error field in the reject response message?

blip-tap.md
ffranr, line 948: I think we should support variable precision. We could add another field to express the precision. For example, the exchange rate could be expressed with these two fields:
```
ScaledExchangeRate uint64
ExchangeRateScalingExponent         uint8
```

`ExchangeRateScalingExponent` is the number of decimal places of precision, the scaling exponent (base 10), which is 4 in the current example.

`ScaledExchangeRate` is equivalent to `suggested_rate_tick`. It's the scaled price of 1 BTC quoted in the counter currency. This is `612,349,500` in the current example.

ffranr: We could use the word `pip` instead, for example:
```
`suggested_scaled_exchange_rate` is the internal representation used for asset
conversions. A pip, in this context, is 1/10000th of a currency unit, providing
up to 4 decimal places of precision (0.0001). For example, if the BTC/USD
exchange rate is $61,234.95, we multiply it by 10,000 to determine the
`scaled_exchange_rate`: `$61,234.95 * 10000 = 612,349,500`.  To convert back to
our normal rate, we divide by `10,000` to arrive back at `$61,234.95`.
```

ffranr: We might want to also consider adding a scaling exponent field (`uint8`) so that we can support variable precision. Otherwise it might be complicated to increase precision in a later release and maintain backwards compatibility.

dstadulis: alternative naming could be `characteristic` https://github.com/lightninglabs/taproot-assets/discussions/763

blip-tap.md
```diff
@@ -0,0 +1,1200 @@
...
+#### Accepting Quotes  (`tap_rfq_accept`) 
+
+If it can, then it should send `tap_rfq_accept` that returns the quote amount
+the edge node is willing to observe to move `N` units of asset `asset_id`:
+
+1. type: ?? (`tap_req_accept`)
+2. data:
+     * [`32*byte`:`rfq_id`]
+     * [`BigSize`:`accepted_rate_tick]
+     * [`BigSize`:`expiry_seconds`]
```
ffranr, line 997: I don't we're representing expiry optimally. Right now, `expiry_seconds` is the number of seconds until the quote expires. But if there's any kind of delay in sending/receiving the message, then that expiry will be interpreted inaccurately. Further, the message wouldn't be "self contained", we need to track the receive time somehow.

I think we should have a field called `expiry_unix_timestamp` with type `uint64`. It's value would be the unix timestamp after which the quote is no longer valid.

blip-tap.md
ffranr, line 1047:
```suggestion
1. type: ?? (`tap_rfq_reject`)
```

blip-tap.md
ffranr, line 996:
```suggestion
     * [`BigSize`:`accepted_rate_tick`]
```

blip-tap.md
ffranr, line 993:
```suggestion
1. type: ?? (`tap_rfq_accept`)
```

PR #30, opened 2023-10-24
bLIP-0029: zero-reserve channels

t-bast:
Nodes can opt-in to remove the channel reserve requirements, which makes better use of their channel liquidity.


blip-0030.md
```diff
@@ -0,0 +1,132 @@
...
+
+When sending or receiving `update_add_htlc`:
+
+* If `zero_reserve` has been negotiated:
+  * MUST ignore any channel reserve standard requirement
+
+If the channel is not public, both nodes:
+
+* When the funding transaction confirms:
+  * MUST send a `channel_update` using the final `short_channel_id`
```
t-bast, line 86: Note this requirement, which is necessary when using `scid_alias`. Without that, the channel updates would only be signed with a local alias, which doesn't allow tying the `channel_update` to a specific on-chain outpoint.

This feels hacky though: we could instead introduce a dedicated proof for the zero-reserve case. This could be as simple as a TLV in `channel_ready` that contains a signature of the funding output's script using the private key associated with the `node_id`.

blip-0030.md
```diff
@@ -0,0 +1,132 @@
...
+
+### TLV extensions
+
+Additional TLV fields for the `open_channel2` message:
+
+1. `tlv_stream`: `open_channel2_tlvs`
+2. types:
+    1. type: 32768 (`zero_reserve`)
+    2. data:
+        * [`byte`:`use_zero_reserve`]
```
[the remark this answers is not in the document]
t-bast: Indeed, we could! Saving 1 byte in a 65kB message isn't a very compelling reason, but being consistent with other similar TLVs is a very good reason. I'll change that.

blip-0030.md
```diff
@@ -0,0 +1,132 @@
...
+  * MAY set the `zero_reserve` TLV field to `1`
+
+When receiving `open_channel2`:
+
+* If `zero_reserve` is set to `1`:
+  * If it wants to use `zero_reserve`:
+    * MUST set the `zero_reserve` TLV field to `1` in `accept_channel2`
+  * Otherwise:
+    * MUST send an `error` and forget the channel
+
```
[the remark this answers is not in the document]
t-bast: Good catch, yes!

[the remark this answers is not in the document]
t-bast: Done in https://github.com/lightning/blips/pull/30/commits/95f7ad10acbb492e2dfe04b5cd06692c0347b9d5

morehouse (approved):
ACK 95f7ad1.

JssDWt:
Does this blip need some info on how to handle dust? The reserve is below the dust bounds, which is unusual. Also there's a possibility to end up with a transaction without outputs in certain scenarios, like when routing the entire channel capacity with dust htlcs.

t-bast:
While it's unusual, it doesn't create specific issues, so we generally don't even need to mention it (outputs are simply trimmed with the existing mechanisms). You have a good point though about a commit tx potentially becoming empty: we must not let that happen!

In practice that should never happen, because:

- there is a limit to dust exposure (https://github.com/lightning/bolts/blob/master/02-peer-protocol.md#bounding-exposure-to-trimmed-in-flight-htlcs-max_dust_htlc_exposure_msat) that should be bounded by a percentage of the channel size
- there is a limit to pending HTLCs (`max_htlc_value_in_flight_msat`) and the recommandation is to set this to something strictly smaller than 50% of the channel capacity

But I agree with you that this should still be explicitly mentioned in this spec. I don't think there's a good way of handling it when it happens, we should rather focus on making sure it doesn't and simply say that you must force-close if the commit tx is becoming empty. Does that sound reasonable @JssDWt?

t-bast:
Sounds good, I'll add that shortly. Thanks!

t-bast:
Done in https://github.com/lightning/blips/pull/30/commits/a912d64c2565623b800bea9201fcaac32f94767c

blip-0030.md
```diff
@@ -0,0 +1,157 @@
...
+If the channel is not public, both nodes:
+
+* When the funding transaction confirms:
+  * MUST send a `channel_update` using the final `short_channel_id`
+
+### Rationale
+
+The use of zero-reserve is symmetrical: it is either offered to both nodes or
+unused.
```
JssDWt, line 108: What is the rationale to make the zero-reserve specification symmetrical? Is that because it's already possible to do one-sided zero reserve?

t-bast: Why wouldn't we make it symmetrical? In order to get the full benefits for a mobile wallet user, you have to allow the LSP side to also be 0-reserve, otherwise you don't benefit from the maximum amount of inbound liquidity.

JssDWt: @t-bast You don't get the full benefit of inbound liquidity, but you do make sure the LSP has something to lose if it tries to cheat. There's value to that.

t-bast: I don't think it actually adds value in practice. In an LSP-wallet relationship, the LSP has enough incentives not to publish a revoked commitment. I detailed those in this thread: https://lists.linuxfoundation.org/pipermail/lightning-dev/2023-October/004136.html

That was the reasoning behind this symmetrical 0-reserve proposal. That doesn't mean you have to use it! If you prefer an asymmetrical proposal, there's always room for another bLIP for that.

niteshbalusu11:
ACK.

t-bast:
I submitted https://github.com/lightning/bolts/pull/1140 to the BOLTs repository: if it is accepted, I will update this bLIP to only contains details for the fraud proof construction (which I don't think makes sense in the BOLTs).

PR #34, opened 2024-07-02
blip-0034: the `recommended_feerates` message

t-bast:
Many lightning messages include feerates for on-chain transactions. Nodes receiving those messages are supposed to reject them if they are considered too low or too high. However, there is no mechanism to let peers know which feerates we currently find acceptable, which leads to frequent failures, especially when the mempool feerates fluctuate quickly.

We introduce an optional message to tell our peers the feerates we'd like to use.

### DIFF
--- a/blip-0034.md
+++ b/blip-0034.md
@@ -1,0 +1,120 @@
+```
+bLIP: 34
+Title: The `recommended_feerates` message
+Status: Active
+Author: Bastien Teinturier <bastien@acinq.fr>
+Created: 2024-07-02
+License: CC0
+```
+
+## Abstract
+
+Many lightning messages include feerates for on-chain transactions. Nodes
+receiving those messages are supposed to reject them if they are considered
+too low or too high. However, there is no mechanism to let peers know which
+feerates we currently find acceptable, which leads to frequent failures,
+especially when the mempool feerates fluctuate quickly.
+
+We introduce an optional message to advertise to our peers the feerates we'd
+like to use when funding new or existing channels.
+
+## Copyright
+
+This bLIP is licensed under the CC0 license.
+
+## Motivation
+
+This feature is particularly useful when mobile wallets initiate channel
+operations with their service provider: in most cases, they should use what
+the service provider recommends since they don't have access to a full node
+and the service provider would reject unacceptable feerates anyway.
+
+## Specification
+
+### The `recommended_feerates` message
+
+1. type: 39409 (`recommended_feerates`)
+2. data:
+   * [`chain_hash`:`chain_hash`]
+   * [`u32`:`funding_feerate_per_kw`]
+   * [`u32`:`commitment_feerate_per_kw`]
+
+1. `tlv_stream`: `recommended_feerates_tlvs`
+2. types:
+    1. type: 1 (`funding_feerate_range`)
+    2. data:
+        * [`u32`:`min_funding_feerate_per_kw`]
+        * [`u32`:`max_funding_feerate_per_kw`]
+    1. type: 3 (`commitment_feerate_range`)
+    2. data:
+        * [`u32`:`min_commitment_feerate_per_kw`]
+        * [`u32`:`max_commitment_feerate_per_kw`]
+
+### Requirements
+
+The sender:
+
+* SHOULD frequently send `recommended_feerates` with feerates it currently
+  accepts for channel funding operations.
+* SHOULD set `funding_feerate_per_kw` to the feerate it uses for funding
+  channels.
+* SHOULD set `funding_feerate_range` to the range of acceptables feerates
+  for funding channels.
+* SHOULD set `commitment_feerate_per_kw` to the feerate it uses for channel
+  commitment transactions.
+* SHOULD set `commitment_feerate_range` to the range of acceptables feerates
+  for channel commitment transactions.
+* SHOULD accept feerates that are inside the ranges it advertised.
+
+The receiver:
+
+* SHOULD not send `open_channel`, `open_channel2`, `splice_init`, `tx_init_rbf`
+  or other channel funding messages using feerates that are outside of the
+  advertised ranges.
+
+### Rationale
+
+While it won't guarantee success every time, taking into account our peer's
+recommended feerates reduces the frequency of failures when funding channels.
+
+### Test vectors
+
+The following test vectors describe how the `recommended_feerates` message is
+encoded:
+
+```json
+[
+  {
+    "recommended_feerates": {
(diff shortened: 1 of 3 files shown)


TheBlueMatt (commented):
I mean, sure, but I'm not really sure how I'd go about implementing this. Pre-Bitcoin Core 28 if I think the fee rate needed for confirmation/mempool acceptance is higher than the channel currently is, I need to update. Whether my peer said they think the fee rates are higher than I think or not I'm probably still gonna use the one I think I need? I guess a peer could tell me they require a fee rate a few % higher and I'd accept that too but the peer should at least accept an increase even if it's not a fee rate they think they need anyway?

More generally, the fix really should be Bitcoin Core 28, anchor channels, and no feerate enforcement.

blip-0034.md
```diff
@@ -0,0 +1,93 @@
...
+
+## Specification
+
+### The `recommended_feerates` message
+
+1. type: 35025 (`recommended_feerates`)
+2. data:
+   * [`chain_hash`:`chain_hash`]
+   * [`u32`:`funding_feerate_per_kw`]
+   * [`u32`:`commitment_feerate_per_kw`]
```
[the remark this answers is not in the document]
t-bast: This is useful when combined with [blip 33](https://github.com/lightning/blips/pull/33), where the non-opener is paying the commit fees: they can tell the opener what feerate they will accept in `open_channel2`.

t-bast:
> More generally, the fix really should be Bitcoin Core 28, anchor channels, and no feerate enforcement.

I don't get what you mean here...sure, for the commitment feerate, this isn't very important once we have anchor channels. But for opening channels and creating splice transactions (funding feerate), you will have a specific confirmation target (especially when combined with 0-conf, which makes RBF impossible).

In the LSP / mobile wallet case, the mobile wallet generally doesn't have access to its own Bitcoin Core instance, and doesn't know what confirmation target the LSP is using. So the LSP just tells the mobile wallet what feerate it will use when it initiates channel-open or splice, and the mobile wallet uses the same values when it is the initiator (e.g. swap-in), which ensures that it will generally be accepted by the LSP.

t-bast:
> Why not send the min/max you'll accept explicitly here?

I hesitated, mostly because the way I'm currently using this, it's not a real guarantee that the recommended feerates will be accepted (depending on how long you wait before initiating an on-chain funding based on those feerates). I felt that having a min/max may give a false assurance that it will be accepted so I decided to stick with the simplest one.

But you're right, it's probably useful to change this to min/max/preferred and keep accepting them until the next `recommended_feerates` is issued, this would be more thorough.

t-bast:
Done and rebased, I hope it's clearer now!

TheBlueMatt (approved):
Without having implemented any of this, makes sense to me.

blip-0034.md
```diff
@@ -0,0 +1,120 @@
...
+Author: Bastien Teinturier <bastien@acinq.fr>
+Created: 2024-07-02
+License: CC0
+```
+
+## Abstract
+
+Many lightning messages include feerates for on-chain transactions. Nodes
+receiving those messages are supposed to reject them if they are considered
+too low or too high. However, there is no mechanism to let peers know which
```
Roasbeef, line 14: > . Nodes receiving those messages are supposed to reject them if they are considered
too low or too high.

We currently don't have any way to reject the fee updates though, other than force closing, which then locks you into a potentially undesirable fee (pre and post anchors, before package relay).

blip-0034.md
```diff
@@ -0,0 +1,120 @@
...
+and the service provider would reject unacceptable feerates anyway.
+
+## Specification
+
+### The `recommended_feerates` message
+
+1. type: 39409 (`recommended_feerates`)
+2. data:
+   * [`chain_hash`:`chain_hash`]
+   * [`u32`:`funding_feerate_per_kw`]
```
Roasbeef, line 39: Should this also be a min+max similar to the `recommended_feerates_tlvs` field below?

Roasbeef: Also I wonder if it's useful for the desired use case if this is always communicating as a mapping from conf target to the fee rate. Otherwise, it's unclear what type of confirmation delay the sender is comfortable with when they send their `funding_feerate_per_kw` field.

blip-0034.md
```diff
@@ -0,0 +1,120 @@
...
+  for funding channels.
+* SHOULD set `commitment_feerate_per_kw` to the feerate it uses for channel
+  commitment transactions.
+* SHOULD set `commitment_feerate_range` to the range of acceptables feerates
+  for channel commitment transactions.
+* SHOULD accept feerates that are inside the ranges it advertised.
+
+The receiver:
+
+* SHOULD not send `open_channel`, `open_channel2`, `splice_init`, `tx_init_rbf`
```
Roasbeef, line 71: Should add `update_fee` as well

blip-0034.md
t-bast: Sure, but what does that have to do with this proposal? This introduction simply states that the BOLTs tell you to reject messages with offending feerates (where we don't really care what "reject" is) but doesn't give you tools to avoid creating a message with an offending feerate, which motivates this bLIP.

blip-0034.md
t-bast: > Should this also be a min+max similar to the recommended_feerates_tlvs field below?

I don't understand what you mean...there is a min/max range defined (`funding_feerate_range`).

> Also I wonder if it's useful for the desired use case if this is always communicating as a mapping from conf target to the fee rate. 

Good question. I don't think it's useful here, since the only thing you want to know is whether your peer will accept the feerate you set, and BOLT messages don't use a conf target but rather feerates.

But it could be informative to have an optional TLV containing the confirmation target the sending node is using, I can add that.

blip-0034.md
t-bast: Done in https://github.com/lightning/blips/pull/34/commits/4cabcbfc0a571c3af610b06606ea4ba53aa89f1c: the commit message explains why `update_fee` is a bit more complex, because it can involve funds safety.

dzdidi:
FYI: @vincenzopalazzo

TheBlueMatt:
Is this still live? Has anchors and zero-fee channels replaced the need for this?

t-bast:
This is still live for Phoenix: we use it to advertise the funding feerate of splice transactions that will guarantee relatively quick confirmation (to avoid creating long chains of unconfirmed 0-conf splices).

It could be simplified now though, since the commitment feerate isn't useful anymore, especially in a 0-fee commitments world.

PR #38, opened 2024-07-03
blip-0038: block header gossip

t-bast:
Lightning nodes must stay in sync with the blockchain to protect themselves against malicious peers that could publish revoked commitment transactions. Honest nodes should help their peers by letting them know about the latest block header they have received. This allows detecting when a node is behind and may need to fix their bitcoin node.

We extend lightning messages to include information about the last header we've received and the corresponding block height.

@Roasbeef can you detail what lnd is doing regarding this feature? You mentioned encoding it in `ping` and `pong`, but what exactly do you include?

### DIFF
--- a/blip-0038.md
+++ b/blip-0038.md
@@ -1,0 +1,80 @@
+```
+bLIP: 38
+Title: Block header gossip
+Status: Active
+Author: Bastien Teinturier <bastien@acinq.fr>
+Created: 2024-07-03
+License: CC0
+```
+
+## Abstract
+
+Lightning nodes must stay in sync with the blockchain to protect themselves
+against malicious peers that could publish revoked commitment transactions.
+Honest nodes should help their peers by letting them know about the latest
+block header they have received: this allows detecting when a node is behind
+and may need to fix their bitcoin node.
+
+## Copyright
+
+This bLIP is licensed under the CC0 license.
+
+## Motivation
+
+We have several messages that are frequently exchanged between lightning nodes.
+They are a good opportunity to include some blockchain data.
+
+## Specification
+
+We add the following TLV field to the `init` message:
+
+1. `tlv_stream`: `init_tlvs`
+2. types:
+    1. type: 32411 (`latest_block_header`)
+    2. data:
+        - [`u32`:`block_height`]
+        - [`80*bytes`:`block_header`]
+
+### Requirements
+
+The sender of `init`:
+
+- SHOULD include the `latest_block_header` field.
+- MUST set `block_header` to the latest valid block header its bitcoin node has received.
+- MUST set `block_height` to the corresponding height in the blockchain.
+
+The reader of `init`:
+
+- If the received `block_height` is greater than the local block height:
+  - SHOULD verify the received `block_header`'s proof of work.
+  - SHOULD alert the node operator if it is missing many blocks.
+
+### Rationale
+
+This extension only allows detecting that we're behind: there is no easy way
+to automatically fix a connection issue to the bitcoin network. Node operators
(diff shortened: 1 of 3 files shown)


t-bast:
> If we're just gonna include this once

I think we'll want to include it in `ping` and `pong` as well, but I didn't do it yet, I'm waiting for @Roasbeef to let me know how they're currently encoding it in lnd first.

But it's probably a good idea to include whatever proof would help to validate the height, I'll need to look into this!

guggero:
The current `ping` payload in `lnd` is the best block header serialized in the default wire format, so 80 bytes: https://github.com/lightningnetwork/lnd/blob/e61cba8d227faa959e1c591cafccf74a8ba98e30/peer/brontide.go#L582

t-bast:
> The current ping payload in lnd is the best block header serialized in the default wire format, so 80 bytes: https://github.com/lightningnetwork/lnd/blob/e61cba8d227faa959e1c591cafccf74a8ba98e30/peer/brontide.go#L582

This is directly encoded inside the `ping` message instead of using a TLV? That doesn't seem very future-proof, it doesn't support padding with random bytes to create `ping` messages of any chosen length? And it doesn't contain any magic byte to indicate that the ping contents contain a block header, so the reader just tries interpreting it as such and ignores it if that doesn't work?

Can you detail why that encoding was chosen? It feels like using an empty `ping` /  `pong` content and using the TLV stream is more flexible, as it allows padding and combining with potentially other data we'd want to exchange in those messages.

guggero:
I would guess that this is a case of "let's just prepare the code to be able to send a custom ping payload but then discuss the actual format at a later point", since we don't currently even look at the payload on the other side, it's just being logged (on `trace` level) and exposed on the RPC: https://github.com/lightningnetwork/lnd/pull/5621

I do agree that using TLV here definitely makes sense.

Roasbeef:
> This is directly encoded inside the ping message instead of using a TLV? That doesn't seem very future-proof, it doesn't support padding with random bytes to create ping messages of any chosen length

Yeah it's just directly encoded, as there's no existing structure defined in the ping messages. Just as the requirement of `pong` is just to send the `N` random bytes. 

> And it doesn't contain any magic byte to indicate that the ping contents contain a block header, so the reader just tries interpreting it as such and ignores it if that doesn't work?

Up until now, there've been no readers 🤷 , they'd identify the payload based on the 80 byte length, and potentially knowledge of the prev block header referenced.

blip-0038.md
```diff
@@ -0,0 +1,80 @@
...
+This bLIP is licensed under the CC0 license.
+
+## Motivation
+
+We have several messages that are frequently exchanged between lightning nodes.
+They are a good opportunity to include some blockchain data.
+
+## Specification
+
+We add the following TLV field to the `init` message:
```
Roasbeef, line 29: An `init` message is only sent once when the connection is established. Putting the payload in `ping` would mean that the new headers can continually be sent and monitored.

t-bast: I see two possible designs for this:

- keep the body of the `ping` messages random and include the block header in the message's tlv stream
- make the body of the `ping` message a TLV stream, that contains the block header (+ padding)

TheBlueMatt: WHy not just a new odd message?

t-bast: We could do that as well: one reason I saw to use existing messages was to fill their content as much as possible to optimize bandwidth usage. If we do message padding, it's better to put actual data in our messages as much as possible rather than random padding?

TheBlueMatt: You don't have to pad with pings, you can also pad with header data, though Rusty's suggested "padding" by simply chunking message sends rather than actually padding also works.

PR #40, opened 2024-07-08
blip-40: Route Blinding Dummy Hops

ellemouton:
This bLIP defines a new TLV field, `next_dummy_priv_key` for the `encrypted_data` 
TLV stream of the onion `payload` type. This new field can be used by the creator of 
a route blinding path to encode the private key to use for peeling off dummy hops from
the onion. This gives a recipient an easy way of identifying when it is the real final hop in 
a path and should therefore continue peeling the onion since only dummy hops remain. 
Using this method, the recipient also does not need to persist any extra data in order to
peel the dummy hops.

### DIFF
--- a/blip-0002.md
+++ b/blip-0002.md
@@ -112,6 +112,17 @@ The following table contains extension tlv fields for the `ping` message:
 |-------|-----------------------------|--------------------------------|
 | 65536 | `tlv_field_name`            | Link to the corresponding bLIP |
 
+
+### `encrypted_data`
+
+The following table contains extension tlv fields for the `encrypted_data` field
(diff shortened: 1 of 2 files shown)


t-bast:
I don't understand why you need a new field for that. Why can't you just encode your own `node_id` as `next_node_id`? That lets you detect that it's a dummy hop, and you perform the standard blinded key derivation based on the current blinding point. And it wouldn't need a bLIP since it's all constructed by the recipient for itself (the writer and the reader are the same)?

ellemouton:
ah yes good point - that should work 👍  Will try that. 

The reason I opened the bLIP for the new field is to reserve that TLV field though incase new fields were added to `encrypted_data`.

PR #41, opened 2024-07-10
blip-0041: channel funding fee credit

t-bast:
Onboarding users to the lightning network is hard, because opening a channel requires paying on-chain fees and allocating capital, hoping that it will be used to earn routing fees. It is not economical nor sustainable to open channels for users that receive a tiny amount and may not use the network frequently enough.

Wallet providers usually work around this issue by requiring users to receive a large enough first payment to pay on-chain fees for their channel.

However, a collection of unrelated tiny payments received over a period of time may be enough to justify creating a channel. We thus introduce a funding fee credit, that lets users accept small payments without creating a channel, and can be used to create a channel once enough payments have been received to cover the on-chain costs.


blip-0041.md
```diff
@@ -0,0 +1,597 @@
...
+  - MUST set `funding_fee` in `update_add_htlc` to collect the fees that weren't paid from fee credit.
+  - MUST set `funding_fee`, even if `fees_msat` is `0 msat`.
+
+### Rationale
+
+The fee credit is always used in combination with on-the-fly funding payment
+types. We collect fees from the fee credit first, which ensures that it
+converges to 0.
+
+The node sending `add_fee_credit` is trusting its peer to correctly keep track
```
Roasbeef, line 131: Can those that accept the free credit use it to make future payments with the credit _without_ an on chain channel? I guess with that it starts to resemble the old "hosted channels" concept.

t-bast: > I guess with that it starts to resemble the old "hosted channels" concept.

That's why we don't want to go down that road...if you need this, just use a custodial wallet. Here we only introduce purchasing on-chain fees in advance, this is purposefully much more restricted to limit the additional trust required.

t-bast:
> There's a lot here, but it seems like we could instead just have a 0conf channel with a fake/unbroadcasted funding transaction, 

It's not just for channel open, this will happen for splices as well. But I did try what you're proposing first after our discussion from the Tuscany Lightning Summit, and once you get down to the nitty-gritty details, it's actually very complex (much more complex IMO than what I'm proposing here).

I'd be happy to be proven wrong, but someone needs to fully implement and specify this alternative protocol to prove that it's simpler and better, and I failed when trying that :sweat_smile:

PR #42, opened 2024-07-19
blip-0042: Bolt 12 Contacts

t-bast:
Bolt 12 introduces offers, which are static lightning "addresses". An offer can be stored and reused to pay the same node many times. It then becomes natural to associate Bolt 12 offers to your friends and contacts.

When sending payments to contacts, you may want them to know that the payment came from you. We propose a scheme to optionally include contact information in outgoing payments to allow the recipient to:

- detect that the payment is coming from one of their known contacts
- otherwise, be able to add the payer to their contacts list
- send funds back to the payer without additional interaction

This feature provides a better UX for lightning wallets, by making payments between contacts look very similar to fiat payment
applications.

### DIFF
--- a/blip-0042.md
+++ b/blip-0042.md
@@ -71,20 +71,32 @@ with their BIP 353 name, thus proving ownership of this BIP 353 name.
 
 #### Requirements
 
+When creating a new *payment contact*:
+
+- MUST receive user authorization to reveal payment information to the contact.
+- If the node has previously received a payment from this contact including the
+  `invreq_contact_secret` field:
+  - MUST associate the received `invreq_contact_secret`, `invreq_payer_offer`
+    and `invreq_payer_bip_353_name` with this contact.
+  - If `invreq_payer_bip_353_name` was included:
+    - SHOULD use it to fetch an `invreq_payer_offer` if none was included.
+    - MUST verify that the received offer matches the `offer_signing_key`
+      provided in `invreq_payer_bip_353_signature`.
+    - SHOULD use it to refresh the `invreq_payer_offer` if it expires.
+    - MAY use it to refresh the `invreq_payer_offer` periodically.
+- Otherwise:
+  - MUST generate a unique `invreq_contact_secret` for that contact, using
+    [deterministic derivation](#deterministic-derivation).
+  - MUST associate this `invreq_contact_secret` with this contact.
+  - MUST include this `invreq_contact_secret` whenever paying this contact.
+
 The writer of `invoice_request`:
 
 - If they want the recipient to be able to identify who paid them:
-  - If the recipient is not yet part of their contacts list:
-    - If they have previously received a payment from this recipient including
-      the `invreq_contact_secret` field:
-      - MUST associate the received `invreq_contact_secret` with this contact.
-      - MUST include this `invreq_contact_secret` whenever paying this contact.
-    - Otherwise:
-      - MUST generate a unique `invreq_contact_secret` for that contact.
-      - MUST associate this `invreq_contact_secret` with this contact.
-      - MUST include this `invreq_contact_secret` whenever paying this contact.
-  - Otherwise:
-    - MUST include the `invreq_contact_secret` associated with this contact.
+  - If the recipient is not a *payment contact*:
+    - MUST seek user authorization to create a *payment contact* matching that
+      recipient.
+  - MUST include the `invreq_contact_secret` associated with this contact.
   - MUST include either `invreq_payer_offer` or `invreq_payer_bip_353_name`.
   - If it includes `invreq_payer_bip_353_name`:
     - MUST set `name` to the post-₿, pre-@ part of the BIP 353 HRN.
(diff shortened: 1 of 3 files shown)


blip-0042.md
```diff
@@ -0,0 +1,71 @@
...
+    - `tweak = SHA256("blip42_bolt12_contacts" || offer.offer_id)`.
+  - MUST choose one of its own offers that the recipient may know:
+    - MUST multiply the private key used to sign invoices for this offer with `tweak`.
+    - MUST use the resulting private key as `payer_key` to sign the `invoice_request`.
+- Otherwise:
+  - SHOULD use a random `payer_key`.
+
+A node receiving `invoice_request`:
+
+- For each `offer` from its trusted contacts:
```
jklein24, line 52: Do you think it might make sense for the payer to include its own offer_id somewhere in the invoice request so that this lookup can be more efficient for the payee?

t-bast: Great idea, that would indeed remove the need for any pre-computation of our contacts `payer_ids` and make it trivial to verify an incoming payment.

t-bast:
> One downside here is it relies on the recipient using a static offer signing pubkey, which by default they almost certainly should not do. A radically different approach would be to use something like bLIP https://github.com/lightning/blips/pull/31 to use separate key material from the offer id to do a mutual handshake without using static keys.

It is very important in my opinion that "adding a contact" is as simple as entering a BIP 353 address or scanning a QR code and clicking a "store contact" button. It is also very important to make it easy to add to any wallet to guarantee compatibility across the ecosystem.

The simplest way I can see to achieve that is to use key material from the Bolt 12 offer directly. This is why I chose to use a key that already needs to be in any valid offer, even though the drawback is that it's a bit of a layering violation (by deriving a key used for _paying_ from a key used to _get paid_).

It seems to me that alternatives will:

- be harder to deploy across the eco-system if we want the key material to be outside of the Bolt 12 offer
- make Bolt 12 offers bigger (which we'd like to avoid) if we add separate key material inside the Bolt 12 offer

t-bast:
> (B) but much more importantly, the use of the offer signing key as the authentication key completely breaks BOLT 12. You must not use the same offer signing key for all of your offers. A static public offer is one thing but for this to work when I scan an offer you also have to reuse the same signing key on every offer, allowing someone to match a fresh offer against my public BIP 353 one.

I don't understand, that's not at all what I'm specifying here? When you'd pay one of your trusted contact, you would use the offer signing key of the current offer you're using in your BIP 353 address (which you can rotate with a different signing key). But previous offers or offers you're using for other purposes have no reason to use the same key?

> I don't think bLIP 32 breaks either of those?

You mean bLIP 31? It doesn't break those assumptions, but I don't see how you fetch/configure the "list of acceptable public keys they wish to exchange messages with" you're mentioning in there if those aren't included in Bolt 12 offers. Or do you mean that you'd like to use bLIP 31, but use the offer signing keys of your contacts as that "list of acceptable public keys" so that they are directly contained in the Bolt 12 offers or your contacts and in their BIP 353 address?

Because if the keys from that "list of acceptable public keys they wish to exchange messages with" does not come from Bolt 12 offers, then we need a new protocol to exchange those keys and somehow include them in BIP 353, which can quickly get messy and hard to get widespread compatibility...

So overall, I'm not at all against using bLIP 31 instead of what I'm proposing here, but I'd like to see a more detailed design that shows how to use it for contact authentication if you could write it down.

t-bast:
> More generally, I may occasionally change my BIP 353 for various reasons, and I don't want my entire contact feature to break just because I changed it? That seems like a weird user footgun.

That's a good point. I'm biased because in Phoenix we deterministically derive a "default" offer from the user's seed (which allows seamlessly restoring it to a different device), but I agree that this is only one out of many possible strategies for mobile wallets, and being able to rotate your default offer without breaking your contact lists is important.

As you note, this does force us to decouple the "contact key" from offers themselves and thus may require re-thinking the UX for adding contacts a bit. Would it make sense to simply include a new "lnck" (LN contact key) field in BIP 21 URIs that contains our contact pubkey? This way we separate it from the offer itself, but our BIP 353 address can include it in the returned URI, and the QR code for receiving Bolt 12 payments shown by wallets could use this BIP 21 URI as well (instead encoding just the Bolt 12 offer).

blip-0042.md
```diff
@@ -0,0 +1,288 @@
...
+   - [`domain_len*byte`:`domain`]
+
+The `invreq_payer_bip_353_signature` field lets payers provide a signature of
+the `invoice_request` using one of the signing keys of the offer associated
+with their BIP 353 name, thus proving ownership of this BIP 353 name.
+
+1. type: 2000001735 (`invreq_payer_bip_353_signature`)
+2. data:
+   - [`point`:`offer_signing_key`]
+   - [`bip340sig`:`signature`]
```
rustyrussell, line 70: Doesn't this signature cover itself?  Which is impossible to implement?  I think this needs to be renumbered to 735?  (Inside the signature TLV range)

blip-0042.md
```diff
@@ -0,0 +1,288 @@
...
+#### Requirements
+
+The writer of `invoice_request`:
+
+- If they want the recipient to be able to identify who paid them:
+  - If the recipient is not yet part of their contacts list:
+    - If they have previously received a payment from this recipient including
+      the `invreq_contact_secret` field:
+      - MUST associate the received `invreq_contact_secret` with this contact.
+      - MUST include this `invreq_contact_secret` whenever paying this contact.
```
rustyrussell, line 81: These requirements are on the reader of an invoice_request, not the writer.

blip-0042.md
rustyrussell: Ah, wait, it uses the BIP353 key.  NVM.

blip-0042.md
```diff
@@ -0,0 +1,288 @@
...
+  - MUST include either `invreq_payer_offer` or `invreq_payer_bip_353_name`.
+  - If it includes `invreq_payer_bip_353_name`:
+    - MUST set `name` to the post-₿, pre-@ part of the BIP 353 HRN.
+    - MUST set `domain` to the post-@ part of the BIP 353 HRN.
+    - MUST include `invreq_payer_bip_353_signature` using the private key of
+      its `offer_issuer_id` or, if its offer doesn't include `offer_issuer_id`,
+      the private key of the `blinded_node_id` of one of its `offer_paths`.
+      The signed data should be computed as detailed in [Bolt 12][bolt12-sig],
+      with "lightning" || "invoice_request" || "invreq_payer_bip_353_signature"
+      as signature tag.
```
rustyrussell, line 97: Doesn't it do this signature already, by signing the invreq itself?

blip-0042.md
rustyrussell: OK, correction.  This should simply say:

```
- If the recipient is a *payment contact*:
      - MUST include the `invreq_contact_secret` whenever paying this contact.
      - MUST include either `invreq_payer_offer` or `invreq_payer_bip_353_name`.
      - ...
```

Then you need a section defining payment contacts:

```
Requirements

When creating a new payment contact:
- MUST receive user authorization to reveal payment information.
- If the node has a previous payment from this recipient including the `invreq_contact_secret` field:
      - MUST associate the received `invreq_contact_secret`, `invreq_payer_offer` and `invreq_bip_353_name` with this contact.
- otherwise:
     - MUST generate an `invreq_contact_secret` using ...
```

Now you can simply refer to this.

There are actually no new requirements on the invoice reader: there are requirements on successfully paying the invoice, so:

```
When an invoice has been paid:
- If `invreq_contact_secret` is present:
   - MAY seek authorization to add a new contact or update an existing contact (if the user verified that the payment came from this contact.)
   - If `invreq_contact_secret` matches an existing contact:
     - SHOULD display the `invreq_payer_note`, if one is provided.
```

Basically, you need to write requirements where they need to be implemented.

blip-0042.md
```diff
@@ -0,0 +1,261 @@
...
+- The private key for the `offer_node_id` is called `offer_priv_key`.
+- When paying `remote_offer` for which we include our `local_offer` in the
+  `invreq_payer_offer` field:
+  - We compute the ECDH of the two `offer_node_id`s:
+    - `shared_key = local_offer.offer_priv_key * remote_offer.offer_node_id`.
+  - We use a tagged hash to derive the `contact_secret`:
+    - `contact_secret = SHA256("blip42_contact_secret" || shared_key)`.
+
+Using this deterministic derivation has multiple benefits. First of all, it
+guarantees that both nodes independently derive the same `contact_secret` when
```
rustyrussell, line 245: Make this a hard requirement please? i.e. put it in the contact generation requirements.

rustyrussell (approved):
I like it, but the requirements sections are tangled.

hMsats:
I would very much like to test this when available in Phoenix

TheBlueMatt (approved):

TheBlueMatt:
If @rustyrussell has any more feedback, its a living document and can be updated.

t-bast:
:tada: :partying_face:

PR #43, opened 2024-08-31
blip-0043: Alternative Addresses

maxrantil:
Alternative Addresses (`alt_addr`) allows nodes to specify private connection addresses for peers with existing relationships. This bypasses the public gossip protocol for reconnections, enhancing privacy and potentially improving reliability and latency.

Nodes can selectively share these alternative addresses with trusted peers, maintaining public presence while offering improved connectivity to specific partners. This feature works alongside existing `alt-bind-addr` and `alt-announce-addr` capabilities, providing more control over network connections and privacy in the Lightning Network.


arbadacarbaYK:
Feel like that would be a nice addition to watchtowers where you already define friendly-node-peers though for a different purpouse atm

PR #54, opened 2024-12-02
Add bLIP 52: JIT Channel Negotiation (LSPS2)

tnull:
Based on #52.

A "JIT Channel" is a channel opened in response to an incoming payment from the public network to a client, via the LSP.
This allows a client with no Lightning channels to start receiving on Lightning, and have the cost of their inbound liquidity be deducted from their first received payment.

The given protocol is based on LSPS0/blip-0050 and has been previously stabilized by the LSP Spec group and is live in production with several LSP and client implementations today.

As previously discussed on multiple occasions, the LSP Spec group is however moving to a bLIP-centric process, which is why we 'upstream' these previously-stabilized specifications here.

(cc the original author @ZmnSCPxj-jr)


ZmnSCPxj-jr:
So.... what is the general process for getting this P.R. merged in?

PR #55, opened 2024-12-02
Add bLIP 55: Webhook Registration (LSPS5) 

tnull:
Based on #52.

In typical mobile environments, a program that is not currently being focused on by the user will be suspended, with all its TCP connections dropped.

This specification provides a way for a mobile client to register some specific webhook by which the LSP can signal a push notification to the application developer server, which will in turn convert the push notification to one that it itself signs and can send to the mobile OS developer server.

The given protocol is based on LSPS0/bLIP 50 and has been previously stabilized by the LSP Spec group.

As previously discussed on multiple occasions, the LSP Spec group is however moving to a bLIP-centric process, which is why we 'upstream' these previously-stabilized specifications here.

(cc the original author @ZmnSCPxj-jr)

### DIFF
--- a/blip-0055.md
+++ b/blip-0055.md
@@ -1,0 +1,647 @@
+```
+bLIP: 55
+Title: LSPS5: Webhook Registration
+Status: Active
+Author: ZmnSCPxj jxPCSnmZ <zmnscpxj@protonmail.com>
+Created: 2024-12-02
+License: MIT
+```
+
+## Abstract
+
+In typical mobile environments, a program that is not currently
+being focused on by the user will be suspended, with all its
+TCP connections dropped.
+
+Typically, such a program would not be able to get either CPU
+time or any ability to receive information from network.
+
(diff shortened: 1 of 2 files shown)


blip-0055.md
```diff
@@ -0,0 +1,662 @@
...
+It takes no parameters `{}`.
+
+`lsps5.list_webhooks` returns an object like the following:
+
+```JSON
+{
+  "app_names": ["My LSPS-Compliant Lightning Wallet", "Another Wallet With The Same Signing Device"],
+  "max_webhooks": 42
+}
+```
```
martinsaposnic, line 273: Is there a particular reason why this doesn't also include the webhook URLs? Including them might make it more ergonomic to work with

martinsaposnic:
> Ah, thanks for pointing that out. (cc @martinsaposnic, mind including that in one of the follow-ups?)

sounds good, will update that today

PR #56, opened 2024-12-02
bLIP-56: Pluggable Channel Factories.

ZmnSCPxj-jr:
Channel factories are a way of scaling Bitcoin by hosting offchain channels, not directly onchain, but in another offchain mechanism.  The hosting mechanism is called a channel factory.  The channel factory can change the funding amounts of channels inside the factory, without any onchain transactions.

This pluggable channel factory protocol is a proposal by which base node software can handle channel state, while letting external code handle state of the channel factory. As far as the node software is concerned, the channel is a "0-conf" channel whose funding transaction never seems to confirm, even though the state is actually backed by an offchain mechanism that ensures that the channel can be unilaterally closed in the worst-case scenario.

When the channel factory wants to change the funding amounts of hosted channels, the node software acts as if a splice operation is being performed, which changes the funding amount (or not changes it, if it is *other* channels in the same factory that are changed but not *this* one). When the channel factory offchain mechanism has settled to a new state, it is as if a splice transaction was successfully deeply confirmed.

### DIFF
--- a/blip-0002.md
+++ b/blip-0002.md
@@ -45,13 +45,14 @@ Custom feature bits used in the `I` [Bolt 11](https://github.com/lightning/bolts
 
 bLIPs may reserve feature bits by adding them to the following table:
 
-| Bits    | Name                  | Description                                                | Context | Dependencies            | Link                      |
-|---------|-----------------------|------------------------------------------------------------|---------|-------------------------|---------------------------|
-| 54/55   | `keysend`             | A form of spontaneous payment                              | N       | `var_onion_optin`       | [bLIP 3](./blip-0003.md)  |
-| 256/257 | `hosted_channels`     | This node accepts requests for hosted channels             | IN      |                         | [bLIP 17](./blip-0017.md) |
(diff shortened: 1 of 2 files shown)


ZmnSCPxj-jr:
Updated blip number to 56, as per PR number.

8144225309:
Currently a small team is building SuperScalar wallet on top of this blip56 design with a cln plugin specific to SuperScalar to keep blip56 neutral and application agnostic.  

Any input, requested changes, feature requests, comments or concerns are greatly appreciated while we are still early in the development process for this bLIP-56 implementation.  Thanks!  

bLIP-56 Fork of Core Lightning:
https://github.com/8144225309/lightning/tree/blip-56
SuperScalar-CLN Plugin:
https://github.com/8144225309/superscalar-cln
WIP SuperScalar Wallet:
https://github.com/8144225309/superscalar-wallet
Original SuperScalar-CLI (implementation reference):
https://github.com/8144225309/SuperScalar
SuperScalar Docs:
https://www.superscalar.win
https://github.com/8144225309/superscalar-docs

PR #57, opened 2025-01-10
bLIP-57: Channel Lease Extensions (LSPS7)

kaloudis:


dzdidi (changes requested):

blip-0057.md
```diff
@@ -0,0 +1,241 @@
...
+- `order_state <string enum>` Current state of the order.
+  - `CREATED` Order has been created. Default value.
+  - `COMPLETED` LSP has published funding transaction.
+  - `FAILED` Order failed.
+- `payment <object>` Contains everything about payments, see [4. Payment](#4-payment).
+- `channel <object>` Contains information about the channel, see [3. Extendable Channel](#3-extendable-channel).
+
+
+**Client**
+- SHOULD validate that `channel_extension_expiry_blocks` + `channel_extension_expiry_blocks` = `new_channel_expiry_blocks`.
```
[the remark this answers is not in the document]
kaloudis: Good catch. Thank you.

blip-0057.md
```diff
@@ -0,0 +1,241 @@
...
+
+
+The request is constructed depending on the client's needs. 
+
+**Request**
+
+```json
+{
+  "short_channel_id": "871428x964x0",
+  "channel_extension_expiry_blocks": 144,
```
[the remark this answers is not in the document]
kaloudis: Good question. Either could work. Open to hear arguments in favor of one over the other.

kaloudis:
We have our service deployed to Mutinynet and testnet3, if anyone would like to poke at it:

Mutinynet: https://mutinynet.com/lightning/node/032ae843e4d7d177f151d021ac8044b0636ec72b1ce3ffcde5c04748db2517ab03

testnet3: https://mempool.space/testnet/lightning/node/03e84a109cd70e57864274932fc87c5e6434c59ebb8e6e7d28532219ba38f7f6df

blip-0057.md
```diff
@@ -0,0 +1,241 @@
...
+License: MIT
+```
+
+# Channel Lease Extensions (LSPS7)
+
+LSPS7 requires [LSPS0](./blip-0050.md) and therefore [BOLT8](https://github.com/lightning/bolts/blob/master/08-transport.md) as a transport layer. The payment object structure is taken from [LSPS1][].
+
+## Motivation
+
+The goal of this specification is to provide a standardized LSP API for wallets to extend the lifetime of a channel purchased from an LSP.
```
joostjager, line 16: Isn't it unfair to pay for the extension of a channel in which the balance has been on the non-LSP side during all of the previous period? No capital cost for the LSP. Or would this be taken into account when pricing the next extension?

I do realize of course that this is more of a remark on the side.

joostjager: Current ratio can of course change drastically, but I was thinking of the average balance of the previous period. So even if the final balance is skewed heavily towards the LSP, it may not mean that the average balance, that is the basis for the calculation of the price of the extension, is too.

A continuous model for this would be to have a constant flow of sats towards the LSP proportional to the LSP balance.

Ofc still side remarks all of this 😅

kaloudis: Interesting idea for sure

MegalithicBTC:
Just wanted to add our two cents here. I definitely understand that some LSPs may want to close channels after a certain period, in order to maintain capital efficiency. For those LSPs it could be useful to have a "renewal" specification like this. 

But: It just seems more natural, from the perspective of a user, that an LSP **never** closes channels, as long as the user is actively using the channel, and bringing their node online at least occasionally (say, every 30 days or so.)  It might not be the most straightforward "return on capital" available for the LSP, but I just think that the latter practice might be better for the user, as well as the network as a whole.

blip-0057.md
```diff
@@ -0,0 +1,241 @@
...
+
+```json
+{
+  "order_id": "bb4b5d0a-8334-49d8-9463-90a6d413af7c",
+  "channel_extension_expiry_blocks": 144,
+  "new_channel_expiry_block": 839374,
+  "token": "",
+  "created_at": "2012-04-23T18:25:43.511Z",
+  "order_state": "CREATED",
+  "payment": {
```
tnull, line 111: This should also support BOLT12 by now, parallel to bLIP-51 (cf. https://github.com/lightning/blips/pull/59).

blip-0057.md
```diff
@@ -0,0 +1,241 @@
...
+```json
+{
+  "short_channel_id": "871428x964x0",
+  "channel_extension_expiry_blocks": 144,
+  "token": "",
+  "refund_onchain_address": "bc1qvmsy0f3yyes6z9jvddk8xqwznndmdwapvrc0xrmhd3vqj5rhdrrq6hz49h"
+}
+```
+
+- `short_channel_id` <[LSPS0.scid][]> Short Channel Identifier (SCID) of the channel to extend.
```
tnull, line 92: I think we need to clarify if this can be any SCID of a previously-purchased channel (maybe even ones towards other peers?), or if it needs to be the actual SCID, or the SCID alias, or, in case of bLIP-52 to `jit_channel_scid`?

blip-0057.md
```diff
@@ -0,0 +1,241 @@
...
+**Client**
+- SHOULD validate that `channel.expiration_block` + `channel_extension_expiry_blocks` = `new_channel_expiry_block`.
+- SHOULD validate the `fee_total_sat` is reasonable.
+- MAY abort the flow here.
+
+**Errors**
+
+| Code   | Message         | Data | Description |
+| ----   | -------         | ----------- | ---- |
+| -32602 | Invalid params  | {"property": %invalid_property%, "message": %human_message% }            | Invalid method parameter(s). |
```
tnull, line 171: Can we leave the invalid property out? Including that might introduce quite a bit of complexity for (IMO) little gain, as it's only human readable. Yes it makes debugging a bit easier, but hopefully that isn't required often?

blip-0057.md
```diff
@@ -0,0 +1,241 @@
...
+
+- LSP MUST validate the order against the options defined in the `lsps7.get_extendable_channels` channel object. LSP MUST return an `100` error in case of a mismatch.
+  - `%option_mismatch_property%` MUST be one of the fields in the `lsps7.get_extendable_channels` channel object.
+  - Example: `{ "property": "max_channel_extension_expiry_blocks" }`.
+
+- LSP MUST validate the request fields. LSP MUST return a `-32602` error in case of an invalid request field.
+  - `%invalid_property%` MUST be one of the fields in the request body. MUST use `.` to separate nested fields.
+  - Example: `{ "property": "channel_extension_expiry_blocks", "message": "Not an integer" }`.
+
+- LSP MUST validate the `token` field and return an error if the token is invalid.
```
tnull, line 184: Probably should mention this is only necessary if the token is actually set.

blip-0057.md
```diff
@@ -0,0 +1,241 @@
...
+- SHOULD validate the `fee_total_sat` is reasonable.
+- MAY abort the flow here.
+
+**Errors**
+
+| Code   | Message         | Data | Description |
+| ----   | -------         | ----------- | ---- |
+| -32602 | Invalid params  | {"property": %invalid_property%, "message": %human_message% }            | Invalid method parameter(s). |
+| 001    | Client rejected | {"message": %human_message% }                                            | [LSPS0.client_rejected_error][] |
+| 100    | Option mismatch | {"property": %option_mismatch_property%, "message": %human_message% }    | The order doesnt match the options defined in `lsps7.get_extendable_channels` channel object. |
```
tnull, line 173: Seems this is missing the 'unrecognized_or_stale_token' error type we have in LSPS1 and 2.

blip-0057.md
```diff
@@ -0,0 +1,241 @@
...
+| ------ | --------- | ------- | ----------------------------------------------------- |
+| 101    | Not found | {}      | Order with the requested order_id has not been found. |
+
+
+### 3. Extendable channel
+
+The `extendable_channel` object returned in `lsps7.get_extendable_channels`, `lsps7.create_order`, and `lsps7.get_order`, has the following properties:
+
+- `original_order <object>` An *optional* property for the original order for the channel lease, including the `id <string>` to identify the order and the `service <string>` the purchase occured on.
+- `extension_order_ids <string[]>` An *optional* list of order IDs for each time the channel lease has been extended.
```
tnull, line 221: I assume this needs to be a subset of `original_order_ids`?

blip-0057.md
```diff
@@ -0,0 +1,241 @@
...
+- `short_channel_id` <[LSPS0.scid][]> Short Channel Identifier (SCID) of the channel.
+- `max_channel_extension_expiry_blocks <uint32>` The maximum number of blocks a channel can be leased for.
+- `expiration_block <uint32>` The block height at which the channel lease will expiry. May be marked as 0 if the user opened the channel to the LSP, and the LSP has no obligation to keep open.
+  - MUST be 0 or greater.
+
+> **Rationale `original_order`** The client MAY want to look up the original order for reference.
```
tnull, line 227: We don't have order_ids in LSPS2, how do we deal with that here?

tnull (commented):
Excuse the delay here! 

Generally makes sense to me to have such a spec, but if we introduce it it might make sense to ensure that it also works for bLIP-52/LSPS2, not only in the bLIP-51/LSPS1 context. Maybe that would require also introducing order ids to LSPS2?

blip-0057.md
```diff
@@ -0,0 +1,241 @@
...
+- `order_id <string>` Id of this specific order.
+  - MUST be unique.
+  - MUST be at most 64 characters long.
+  - SHOULD be a valid [UUID version 4](https://en.wikipedia.org/wiki/Universally_unique_identifier#Version_4_(random)) (aka random UUID).
+- `funding_confirms_within_blocks <uint16>` Mirrored from the request.
+- `token <string>` Mirrored from the request.
+  - MUST be an empty string if the token was not provided.
+- `created_at` <[LSPS0.datetime][]> Datetime when the order was created.
+- `order_state <string enum>` Current state of the order.
+  - `CREATED` Order has been created. Default value.
```
tnull, line 155: Why do we need to publish another funding if we extend the channel?

blip-0057.md
```diff
@@ -0,0 +1,241 @@
...
+  - `COMPLETED` LSP has published funding transaction.
+  - `FAILED` Order failed.
```
tnull, line 157: How do we imagine the user gets reimbursed if they pay via Lightning? Do we still need to clarify that the LSP has to hold the HTLC until they updated the order state fully to make the state transition atomic (although we probably don't need a HOLD state)?

tnull (commented):
Some more comments. It might also be good to think this through in conjunction with splicing (cf. https://github.com/lightning/blips/pull/65 for example). In particular, it could get messy which 'parts' of the balance are considered expired when exactly, for example.

PR #61, opened 2025-04-29
fix: link to Torv3 hidden service specification

sheroz:
The old link: [A Torv3 hidden service](https://gitweb.torproject.org/torspec.git/tree/proposals/224-rend-spec-ng.txt) does not exist


vincenzopalazzo (approved):
LGMT

PR #69, opened 2026-03-08
LSPS: Add optional client_info field for client and node identification

kaloudis:
## Summary

- Adds an optional `client_info` object to LSPS0 Common Schemas (bLIP 50) for identifying client applications and node implementations
- References `client_info` from `lsps1.create_order` (bLIP 51) and `lsps2.get_info` (bLIP 52)
- All fields are optional strings (max 64 bytes): `app`, `app_version`, `node`, `node_version`

## Motivation

LSPs currently have no standardized way to know which client applications or node implementations are using their services. Some implementations work around this by overloading the `token` field, but this conflicts with its intended purpose (discount codes, API keys). A dedicated `client_info` object cleanly separates client identification from authentication/authorization, enabling better debugging and analytics without sacrificing the `token` field.

## Design Decisions

- **Defined in LSPS0** rather than per-spec to avoid duplication — any LSPS method can reference it
- **All fields optional** — clients can send as much or as little as they want
- **Free-form strings** — no enforced casing or versioning format (e.g. `"0.13.0-beta1"` is valid)
- **LSPs MUST NOT require it** and **MUST NOT use it for discriminatory behavior**

### DIFF
--- a/blip-0050.md
+++ b/blip-0050.md
@@ -1073,6 +1073,46 @@ Example:
 }
 ```
 
+### Client Info
+
+###### Link: LSPS0.client_info
+
+An optional object that allows the client to identify its application
+and node implementation to the LSP.
+This can be used by LSPs for analytics, debugging, and compatibility
+purposes.
+
+All fields are optional JSON strings of at most 64 bytes each.
+The client MAY omit any or all fields.
+The client MAY omit the entire `client_info` object.
+
+The LSP MUST NOT require `client_info` to be present.
+The LSP MUST NOT use `client_info` to discriminate against clients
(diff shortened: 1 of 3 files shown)


tnull:
Hmm, to be honest I'm not super convinced we should do this, as it definitely *has* drawbacks w.r.t. privacy and introduces the risk of being abused as a form of 'version negotiation' field, i.e., services might use this to opt into a proprietary superset of the spec, leading to even more fragmentation across the ecosystem.

> MUST NOT use it for discriminatory behavior

Do you believe this really won't happen once this is introduced?

PR #71, opened 2026-05-14
Add error failures to DNSSEC lookups

TheBlueMatt:
Not sure really why these weren't included, whether they can be trusted or not its very useful to have a resolver indicate you shouldn't wait around any longer for a response.

### DIFF
--- a/blip-0032.md
+++ b/blip-0032.md
@@ -32,6 +32,12 @@ Two new onion messages are defined, `dnssec_query` and `dnssec_proof`.
     * [`u16`:`proof_len`]
     * [`proof_len*byte`:`proof`]
 
+1. type: 65550 (`dnssec_error`)
+2. data:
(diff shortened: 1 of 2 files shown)


blip-0032.md
```diff
@@ -32,6 +32,11 @@ Two new onion messages are defined, `dnssec_query` and `dnssec_proof`.
     * [`u16`:`proof_len`]
     * [`proof_len*byte`:`proof`]
 
+1. type: 65550 (`dnssec_error`)
+2. data:
+    * [`u8`:`domain_name_len`]
+    * [`domain_name_len*byte`:`domain_name`]
```
benthecarman, line 38: should error be included? would be worthwhile to know if the dns entry doesn't exist vs the node having an error

benthecarman (approved):

vincenzopalazzo (approved):
LGTM

